PR #69, opened 2013-05-14
Binary-safe Response and Requests

bmharper:
Many internal buffers have been changed from String to
byte[]. Wherever we need to convert between the two, we
use the function SparkUtils.stringToBytes or
SparkUtils.bytesToString, which uses UTF-8 always.

In order to read the request body as a byte[], one uses
Request.bodyBytes().

In order to write a response as a byte[], one simply
returns a byte[] from the handler.

All the existing tests pass without alteration,
so I hope this doesn't break existing code.

There is one new test which demonstrates the binary-safe IO.

### DIFF
--- a/src/main/java/spark/Access.java
+++ b/src/main/java/spark/Access.java
@@ -20,8 +20,8 @@ public final class Access {
 
     private Access() {}
     
-    public static String getBody(Response response) {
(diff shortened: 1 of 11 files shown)


perwendel:
Rejected.

PR #121, opened 2014-01-17
Updated pom.xml to build an OSGi bundle instead of a plain jar.

steve-taylor:
I want to use Spark in OSGi, but it isn't an OSGi bundle yet. All its dependencies are OSGi bundles. This change simply makes Spark build as an OSGi bundle, which is exactly the same as before, but with some extra OSGi-specific information in META-INF/MANIFEST.MF. No change is required to the way you build. Simply running `mvn install` will produce the bundle and it will have the same file name as before.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,193 @@
-   Copyright 2011 Per Wendel
+
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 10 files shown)


perwendel:
duplicate of #244

PR #130, opened 2014-03-22
Add toString to Request

InfernoZeus:
Needed this for my own debugging purposes, hope it's useful for others.

This pull request also includes the changes in #129 , as it currently depends on the uri() and version() methods I added. I made the other pull request in case you don't want to include the toString.


perwendel:
Can break backwards compatibility for users depending on "old" toString functionality.

PR #132, opened 2014-03-23
Update Jetty to version 9.1.3

InfernoZeus:
Includes a couple of changes to deal with any API differences.


perwendel:
Not working. Updates need to be made to JettyHandler etc.

Issue #85, opened 2013-08-02
Using staticFileLocation("/public") in a project (Using version spark-core-0.9.9.7-SNAPSHOT) deployed on an external server does not work. 


maagie:
Version spark-core-0.9.9.7-SNAPSHOT
Was trying to use staticFileLocation("/public") in my code as follows:

public class MyClass implements spark.servlet.SparkApplication{

```
@Override
public void init() {
    staticFileLocation("/public");
            ........
            ........
        }
```

When I generate a war for the project and deploy it on Jetty I get the following exception:
java.lang.IllegalStateException: This must be done before route mapping has begun
        at spark.Spark.throwBeforeRouteMappingException(Spark.java:320)
        at spark.Spark.staticFileLocation(Spark.java:144)
        at com.itgssi.pmds.client.pam.MyClass.init(MyClass.java:27)
        at spark.servlet.SparkFilter.init(SparkFilter.java:59)
        at org.eclipse.jetty.servlet.FilterHolder.initialize(FilterHolder.java:135)
        at org.eclipse.jetty.servlet.ServletHandler.initialize(ServletHandler.java:800)

This is happening because The SparkFilter I define in my web.xml as follows:

```
                                                   <filter-mapping>
                                                           <filter-name>SparkFilter</filter-name>
                                                           <url-pattern>/pam</url-pattern>
                                                   </filter-mapping>
```

is initializing the servlet and is not waiting for the SparkApplication to Load. 
Hence the servlet is initialised even before the class is loaded, causing the exception.
This needs to be resolved as one cannot deploy a class using staticFileLocation on an external webserver.

PR #143, opened 2014-05-18
Fix for #85. Static resource when run in Tomcat (or other app server)

perwendel:
Fix for #85 
Please review this fix before I'll merge it.

### DIFF
--- a/src/main/java/spark/resource/ExternalResourceHandler.java
+++ b/src/main/java/spark/resource/ExternalResourceHandler.java
@@ -20,15 +20,17 @@ package spark.resource;
 import java.net.MalformedURLException;
 
 import org.eclipse.jetty.util.URIUtil;
+import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


Entwanderer:
I'm using spark with tomcat8. 
```javascript
staticFileLocation("/public")
```
seemed to work.

But there is a problem, all the files are served with content type "text/plain".
So when including CSS files with absolute path
```javascript
<link rel="stylesheet" href="/example.css"/>
```

it will work and be used in FireFox and Chrome. But Chrome throws an error that it must be "text/css". And Internet-Explorer (even the newest Edgte-Version) does also so, and is even not loading the CSS!

See SEC7113: "SEC7113 - CSS was ignored due to mime type mismatch"
https://msdn.microsoft.com/en-us/library/hh180764%28v=vs.85%29.aspx

Is there already a way to activate/configure Content-Types/MIME for static files... ?

PR #164, opened 2014-06-28
Build spark as OSGi bundle.

davsclaus:
This little change to the pom.xml builds the JAR as an OSGi bundle so people can deploy spark in OSGi containers.


perwendel:
duplicate of #244

PR #167, opened 2014-07-01
Implemented instance API as an alternative to the static API.

ggalmazor:
This has to do with prior pull-request #77. I tried to solve the conflicts for that pull-request but finally decided to start form scratch:

  - The static API remains untouched.
  - Migrated logic to SparkInstace class
  - Replaced logic from Spark & SparkBase to use a default instance of SparkInstance
  - Written instance API versions of all tests
  - Written a test to probe that we are able to run two instances at the same time on different ports

Instances of SparkInstance are obtained through a Builder class. This enables to get instances in a much safer way, as setters are now protected (no risk of IllegalStateException when using instances).

Take a look at [InstanceBooks.java](https://github.com/ggalmazor/spark/blob/master/src/test/java/spark/examples/books/InstanceBooks.java) and all the tests in [src/test/java/spark/instance_api folder](https://github.com/ggalmazor/spark/tree/master/src/test/java/spark/instance_api) for usage examples of the new API.


borud:
Has there been any progress on this?  Has @perwendel voiced any opinion on whether he will eventually want to merge in this change?  Right now, this is probably the most important reason I am hesitant to use Spark.

PR #185, opened 2014-08-17
Make it possible to use Spark 2 with Java 7 again

kroepke:
My projects cannot go to Java 8 yet, but still I would like to use Spark 2 as I believe Spark 1.1 will no longer see updates.
The changes necessary to make it compatible are trivial, without affecting the ability to use Lambdas in Java 8 code.
Please have a look at it, I'd be happy to donate the changes, but I realize that this change affects the setup significantly.

Commit message below:

move tests to a separate module to be able to run it with a different language level within IDEA
 - com.sparkjava:tests runs on java 8 to make use of lambdas
 - IDEA does not support different language levels for code and tests, so it has to be in separate modules

changes to spark-core are trivial, just making some variables final as per Java7 requirements


CrisLi:
+1

nchiring:
+1

yadongwen:
+10086

tipsy:
This is almost two years old, it's realistically not going to happen... Hopefully there are fewer people locked to Java 7 out there now.

Issue #200, opened 2014-10-05
Please hide Jetty version in response headers by default


ngocdaothanh:
Also, with the current version of Spark, is there a way to hide Jetty version in response headers?

c-rack:
@ngocdaothanh yes, you could simply override it with a filter like this:

```java
before((request, response) -> {
    response.header("Server", "Jetty");
});
```

tipsy:
Duplicated by #331

PR #212, opened 2014-11-17
Hide Jetty version in response headers by default #200

gutobortolozzo:
Providing HttpConnectionFactory injection point to configure embedded jetty.

### DIFF
--- a/src/main/java/spark/Spark.java
+++ b/src/main/java/spark/Spark.java
@@ -16,6 +16,10 @@
  */
 package spark;
 
+import static java.util.Objects.requireNonNull;
+
+import org.eclipse.jetty.server.HttpConnectionFactory;
+
 import spark.exception.ExceptionHandlerImpl;
 import spark.exception.ExceptionMapper;
 import spark.route.HttpMethod;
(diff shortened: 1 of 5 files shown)


tipsy:
@gutobortolozzo The title doesn't seem to reflect the code changes. What is this really supposed to do?

gutobortolozzo:
@tipsy It was supposed to allow `HttpConnectionFactory` to be configurable and allow developers to call something like `httpConnectionFactoryfactory.getHttpConfiguration().setSendServerVersion(false);`

tipsy:
Oh, I see. It's really that complicated to remove the header, huh.. Would it be simpler to just set `setSendServerVersion(false)` directly instead of allowing devs to configure it?

gutobortolozzo:
@tipsy It is one possible approach. My idea to expose `HttpConnectionFactory` came mainly because `HttpConfiguration` has several other configuration methods that would be nice to expose but add it to Spark seems like a burden, IMHO.

tipsy:
> What is your opinion about this?
I think you make a very valid point. We want to keep Spark's API as simple as possible, so doing configuration through `HttpConfiguration` seems to be a lot better than exposing individual methods. This PR needs to be renamed though. If you update it I'll see if I grab @perwendel for a quick chat and merge it.

tipsy:
> @perwendel says:
> For the "main" API (Spark class) we should have generic and often used config, like threadPool etc.
> Other web-server specific config should be done via EmbeddedServers (which is not yet ready for this). Ie, they would have to provide configuration to the current used embedded server or the one they want to change it with. e.g.
> `EmbeddedJettyServerFactory.configuration(...)`

So, we can't use this approach. Hm. Keeping the PR open though.

gutobortolozzo:
@tipsy I will take a look at those classes and see how I can configure Jetty using them to post a new PR.

tipsy:
This should be taken care of by #813

Issue #54, opened 2013-03-28
Custom 404 Page when using Static File Route?


heller:
When using `staticFileRoute("/foo")`, is there an easy way to configure a static, catch-all 404 error page? (Defining a `get(new Route("*"){…})` at the end causes the static files to not be served.)

mnrasul:
+1
Even if there is a special route for errors, it would be useful. Right now, there is no way to handle any error messages and staticFileRoutes at the same time.

Thanks.

draeron:
started using spark framework 2 weeks ago, about and this is the first issue i ran into

mnrasul:
I looked under the hood, and there is no reasonable way to fix this. Perhaps, I couldn't understand it.

Either ways, its telling that this has been open for 2 years, and I expect it to be a basic feature.

I'm tinkering with Dropwizard.io now and like it so far, perhaps, you will consider it worthwhile as well.

nkavian:
+1

perwendel:
Implementation of static resources changed. This is now fixed.

perwendel:
re-opening until PR merged...

PR #213, opened 2014-11-17
Adding 404 and 500 custom page error

debarshri:
Added a ExceptionUtil for adding custom 404 and 500 error pages.

### DIFF
--- a/400.html
+++ b/400.html
@@ -1,0 +1,1 @@
+<html><body>Yo! yo! custom error..Hello Static World!!!!</body></html>
--- a/500.html
+++ b/500.html
@@ -1,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<head lang="en">
+    <meta charset="UTF-8">
+    <title></title>
+</head>
+<body>
+Yo! yo! custom 500 error..Hello Static World!
+</body>
+</html>
--- a/pom.xml
+++ b/pom.xml
@@ -85,6 +85,11 @@
             <version>2.2.4</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.4</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/src/main/java/spark/Spark.java
+++ b/src/main/java/spark/Spark.java
@@ -19,6 +19,7 @@ package spark;
 import spark.exception.ExceptionHandlerImpl;
 import spark.exception.ExceptionMapper;
 import spark.route.HttpMethod;
+import spark.utils.ExceptionUtils;
 import spark.utils.SparkUtils;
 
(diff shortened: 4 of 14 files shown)


martylamb:
This is nice.  I'm not yet familiar with this code flow but is there any reason not to provide arbitrary error page handlers rather than specific 404 and 500 handlers?

something like:

```java
    // other spark setup
   halt(500, new My500HaltHandler());
   halt(404, new NotFoundHandler());
   halt(new DefaultErrorHandler()); // is used if no status-code specific handler is specified
```

debarshri:
It has been pretty longtime..I even forgot what it was actually about..Let me take alook!

debarshri:
So..the thing that I have tried to solve is when a you have a route which is not specified in the list of routes..it redirects to a error page. I think you cannot do that with halt. 

To be more specific.

Lets assume you have declared routes /route1 and /route2, but client requests for /route3, then it should be redirected to error pages.

It is related to issue https://github.com/perwendel/spark/issues/54

mnrasul:
Good effort. Thanks. 
The thing is there are all kinds of errors which one would like to handle. The current implementation is fairly restrictive. Can it be made more generalized?

OzzyTheGiant:
To @mnrasul and anyone who wants to handle multiple errors at once, my approach was to do a catch all exception function:

```java
        // all other code here
       
       exception(Exception.class, exceptionHandler);
}

public static ExceptionHandler<Exception> exceptionHandler = (exception, request, response) -> {
	var statusCode = 500; // default status code
	var errorMessage = exception.getMessage(); // default error message to output
	
	if (exception instanceof HttpException) {
		HttpException httpException = (HttpException) exception;
		statusCode = httpException.getStatusCode();
		errorMessage = httpException.getMessage();
	}
	response.status(statusCode);
	response.body(String.format("{\"message\": \"%1$s\"}", errorMessage));
};
```

 I created a generic HttpException class for all http errors that can occur in app (400, 404, 500, etc.)
 Each http exception will have it's own sub class (example: BadRequestError for 400), although you may want to name them something more specific for logging purposes, but when instantiating a new exception, the constructor will call super() for HttpException with the message and a status code. This is done so that you don't have to do an `instanceof` check for every error class, just for HttpException. If the exception in question is not an instance of HttpException, it will simply just get the message from that exception class. The HttpException class could look like this:

```java
import java.io.IOException;

public class HttpException extends IOException {
	private static final long serialVersionUID = 1L;
	private static String defaultMessage = "Server Error: Please try again";
	protected int code = 500;

	public HttpException() {
		super(defaultMessage);
	}

	public HttpException(String message) {
		super(message);
	}

	public HttpException(String message, int code) {
		super(message);
		this.code = code;
	}

	public int getStatusCode() {
		return this.code;
	}
}
```

Perhaps this could be added either to docs or as part of a tutorial in the Tutorials web page

Issue #195, opened 2014-09-17
Enhancement: concept of "finally" filter


revetkn:
I noticed that "after" filter methods are not invoked if an exception occurs, making it possible to have mismatched "before"/"after" pairs.  Suppose I begin a database transaction in "before" and an exception occurs - the "after" does not fire, so it is up so some unrelated place like ExceptionMapper to perform cleanup.

I confirmed this by examining https://github.com/perwendel/spark/blob/master/src/main/java/spark/webserver/MatcherFilter.java

What do you think of introducing a "finally" filter, which would always execute, even in the face of an exception?  If you think it makes sense, I can probably come up with a pull request.  Or maybe there is a better way to do this out-of-the-box that I am missing and we don't need any Spark changes at all :)

chillenious:
Just ran into this while going after a bug. I actually expected the 'after' filters to be executed regardless. Anyway, whether explicit or implicit, I think it is important that you can define filters that are always ran, exception or not.

gutobortolozzo:
Hi @chillenious and @perwendel, a few months ago I submited the PR #217, can you take look and provide some feedback about the feature? 
When I needed this feature the code submited worked for me. Maybe it help you, and can be merged.

chillenious:
Looked at it, but I'm not a developer on this project. Imho exceptions - in
particular because they are being used internally (halt) - should not block
exceptions. I'm using filters to clean up any thread locals, set headers I
always want to set regardless of the return, etc, and I had to implement a
pretty gnarly workaround to get that done.

Also, I don't think the before/ after filters are quite as convenient as
they first come across when compared to filters, as I often have the case
where I want to do something before and after, which now takes me two
filters and probably some temporary shared state to accomplish. Maybe
something to reconsider for a 2.0 version? :-)

MouettE-SC:
PR has been merged, this issues can be closed now.

PR #217, opened 2014-11-29
Enhancement: concept of "finally" filter 

gutobortolozzo:
It runs in the finally clause of MatcherFilter, so it give you the guarantee of the filter will be executed even if exception is thrown.

Can you provide me some feedback about this code, and if its contemplate the whole idea discussed on issue #195 ?

### DIFF
--- a/src/main/java/spark/FilterImpl.java
+++ b/src/main/java/spark/FilterImpl.java
@@ -72,5 +72,4 @@ public abstract class FilterImpl implements Filter {
     String getPath() {
         return this.path;
     }
-
 }
--- a/src/main/java/spark/FinallyFilter.java
+++ b/src/main/java/spark/FinallyFilter.java
@@ -1,0 +1,14 @@
+package spark;
+
+/**
+ * Markup abstract class
+ * 
+ * @author Bortolozzo
+ *
+ */
+public abstract class FinallyFilter extends FilterImpl {
+
+	public FinallyFilter(String path, String acceptType) {
+		super(path, acceptType);
+	}
+}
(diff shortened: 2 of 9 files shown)


kodfodrasz:
This one is needed badly!

basdek:
I agree with @kodfodrasz that we need this one badly (or anything that can provide us with such a thing, perhaps another solution is more middleware like exception handling). 

I can see something like this work:

```

Spark.exception(Exception.class, (e, req, res) -> {
 //do all things finally / transactiony;
}

Spark.exception(MySpecificExceptionThatNeedsSpecificResponseHandling.class, (e,req,res) -> {
}

```

kodfodrasz:
@basdek  I think a finally clause would be better, as it would be idiomatic with the java exception handling mechanism, and because it would enable more concise code, and it wouldn't need the knowledge of Spark internals (like halt is handled by a special exception type).

basdek:
@perwendel  @gutobortolozzo How can we help fixing this? And if @gutobortolozzo, is not able to finish it, can we continue working this PReq, perhaps via a fork of your repo?

gutobortolozzo:
@basdek @kodfodrasz I willing to start working again in this PR, since more people need it, probably tomorrow I will take a look at my one year old code and see if it keep working after a merge with master code :) 
If you have any suggestion, it would be welcome.

@perwendel Any thoughts about this feature? What's your opinion about a finally filter?

kodfodrasz:
great news! I'm looking forward to having it in spark! :+1:

gutobortolozzo:
#406

basdek:
@gutobortolozzo Thank you for this awesome work man! (Will be checking into it as soon as my meetings are over today)

Issue #201, opened 2014-10-09
Request#body() empty in Route if previously accessed in a Filter


mikolak-net:
If `Request#body()` is called in a `Filter` mapped for a HTTP request, subsequent calls in the `Route` for this request will return an empty string.

I suspect this is simply to due to a new `Request` being created for each processing stage, and the previous read exhausting the corresponding `InputStream` in the `HttpServletRequest`.

ecamaj:
This is not working in "before" filters too. After filter is working OK, but in before, req.body() is lost after first "consummation".

Frietziek:
Same problem here with before and body coming with an empty string.

Issue #202, opened 2014-10-09
`Request#queryParams` slurps up body if called first


mikolak-net:
Related issue to #201 : when calling `Request#queryParams` _the entirety of the request body becomes a key_ in the query param map, and `Request#body()` returns an empty string. This happens even if within a single processing unit (`Route`/`Filter`).

OptimusV5:
I meet this issue in the newest version 2.7.2 @perwendel help please

PR #218, opened 2014-12-04
Fix for #201, making body available.

perwendel:
Changes made in MatcherFilter and request classes to avoid re-creating request instances by just changing the match related.


valtterip:
We ran into this issue with Spark 2.7.2. The data was gone by just debug-printing the request.body(). Someone commented on issue #202 they saw it with 2.7.2 as well. Regression? @perwendel please consider re-opening. Thanks.

PR #222, opened 2014-12-12
Makes Jetty's thread pool configurable.

kumai:
Current SparkServer uses default settings of Jetty's thread pool, and those are not configurable. I think those settings should be configurable for high load condition.


perwendel:
I'll do some updates to the code.

perwendel:
I did fixes to this code in #273. It would be nice if @kumai could review my changes. For future reference, all public methods needs correct javadoc. If not maven central won't accept the release.

PR #232, opened 2015-01-27
Fixed #231: Added support for removing all routes and particular ones

### LINKED ISSUE #231: Remove all or certain routes
We have a need to remove all routes and certain routes (given a path and an optional HTTP method) from our class that derives from `SparkBase` without having to restart the Web server. To support this, `SparkBase` should expose the following methods:

```
+ removeRoute(path: String): boolean
+ removeRoute(path: String, httpMethod: HttpMethod): boolean
+ clearRoutes(): void
```

These would then forward the request to `SimpleRouteMatcher` which needs to expose analogous methods.

travisspencer:
Here's a fix for issue #231, including docs and unit tests.


AUTplayed:
can this still be used today? because I can't seem to find out how...

PR #235, opened 2015-02-01
Problem: Static files are not served binary-safe. Fixes #215.

### LINKED ISSUE #215: staticFileLocation messes up binary files (e.g. images)
In SparkFilter>>doFilter() there is an IOUtils.copy that copies from an InputStream to a Writer with no encoding. If the resource is a binary file (e.g. png) then its content will be corrupt.

I propose changing that line to: IOUtils.copy(resource.getInputStream(), response.getOutputStream());

What do you think?

Regards
Attila Magyar

c-rack:


avarabyeu:
Guys, i'm very interested in this fix. Are there any plans on pushing this to master and releasing spark 2.2?

c-rack:
@avarabyeu Thanks for your comment. I have contacted Per by mail 3 weeks ago, but without response. Last information is his [tweet](https://twitter.com/perwendel/status/559724862860984320) from 2015-01-26:
> "Wonder why so little currently happens with Spark? On parental leave with two kids. Spark stuff will happen in a couple of weeks"

Spark is such an awesome framework and there are so many people offering help and trying to improve it (currently there are 30 open pull requests), so it would be great if @perwendel could give another update about his Spark roadmap.

Until then, you could only build your own version with the fix included and use it as a local JAR. If you have a local [Nexus](http://www.sonatype.org/nexus/) instance, that shouldn't be that complicated. If you do not have Nexus, this [answer on Stackoverflow](https://stackoverflow.com/a/4955695) might help.

avarabyeu:
@c-rack Thanks for the update! Building local JAR is not a problem, but, anyway, it will be good to have this stuff merged to master since bug is pretty critical.  
As for spark, i was wondering if @perwendel may grant merge rights to some trusted people in order to keep framework and community alive :)
P.S. thanks for the fix!

benjyz:
@c-rack I agree about merging rights. don't know if there are several active /trusted contributors to do this. I'm switching to a local build at the moment.

perwendel:
@c-rack @benjyz @avarabyeu I would like to give merging rights to people i would trust and I'm constantly trying to find the right person(s) for this. I can't give a roadmap yet since I just don't know when I have the time to release a 2.2. I'm home with two kids and at the same time I get many pull requests each weeks and many emails a day so the work load is to high to keep up with. I'm very greatful for all contributions and that I have been able to give the webpage and documentation responsibilty to a very skilled guy in Norway. I'm currently looking through the PRs to figure out which have high prio and which have not.

perwendel:
:+1:

Issue #229, opened 2015-01-12
Gzip compression of content


fab1an:
I found no way of doing this in Spark. In Jetty you can do it this way:

```
GzipHandler gzip = new GzipHandler();
gzip.setMimeTypes("text/html");
gzip.setHandler(new HelloHandler());
```

pazhapn:
+1

gencube:
How about this???
http://sparkjava.com/documentation.html#response-transformer
```java
get("/hello", "text/html", (request, response) -> {
    return new MyMessage("Hello World");
}, new GzipTransformer()); // 
```

tipsy:
This should work automatically now: 

    after((request, response) -> {
        response.header("Content-Encoding", "gzip");
    });

PR #237, opened 2015-02-04
Adding gzip compression support

EdwardRaff:
Modifying the JettyHandler to use GZIP compression when the client
indicates support. Also includes an utility wrapper for implementing the
ResponseWrapper. This probably still needs more work to fit in with the
framework / be configurable. 

I don't normally work with any web frameworks, so any direction on how I could clean this up so it can be added would be appreciated.


EdwardRaff:
I was looking at this for a potential future project, and gzip would be a needed feature for me. So I figured I would give a shot at it (also in reference to issue #229 )

jkwatson:
This is a great change. :+1:

EdwardRaff:
Ok, after playing around I changed the way its implemented (no longer needs extra helper classes). 

Unfortunately, the use of strings for body makes things a little more complicated than they should be - It would seem better if I could have a filter take an input stream and then return a new stream. Then have a the default case be a String returned (wrapped to a output stream). But any way. 

New code is done in the MatcherFilter such that it checks for the *Content-Encoding:gzip* header, and if present, wraps the output stream in a gzip stream (closed by try-catch, I'm not sure if thats a problem since the current implementation doesn't close the stream it grabs in the default case). I can now do something like 

```java
after("/dataDownload/*", "application/json", (req, resp) ->
{
   resp.header("Content-Encoding", "gzip");
});
```

To selectively apply gzip compression only to the json requests under dataDownload (or whatever other combination someone might want) 

Just to make sure, I tested it in Chrome with 
```java
static int count = 0;
public static void main(String[] args) 
{
    get("/hello", (request, response) -> 
    {
        StringBuilder sb = new StringBuilder();
        for(int i = 0; i < 100000; i++)
            sb.append("Hello World").append("\n");
        return sb;
    });
        
    after("/*", (req, resp) ->
    {
        if (count++ % 2 == 0)
            resp.header("Content-Encoding", "gzip");
    });
}
```

It switched between downloading 1.1MB of data over 331 ms (no gzip) to downloading 2.5KB over 11 ms.

perwendel:
Nice!
Since #240 has been merged I guess this needs some updates?

EdwardRaff:
Yes, it will - but #240 is a very good change. 

I'll sit down soon and update the code. Does the style of enabling gzip that I used make sense? Or should I change that while updating?

perwendel:
I think it makes sense. Not my strongest area though.
I'm ramping up for a 2.2 release very soon (this week). I'd really like to have this included if you have the possibility to update this that soon.

EdwardRaff:
I think so, though you give me little time! 

It looks like there is a bug in the #240 pull. Treating the bytes as a string when handing an InputStream has the potential to causes issues when deciding - especially if it's not a string. That is the code that I would be exercising with gzip support.

perwendel:
@EdwardRaff You mean the code in the InputStreamSerializer?

Would it be better doing:

```java
@Override
public void process(OutputStream outputStream, Object element)
       throws IOException {
    InputStream inputStream = (InputStream) element;
    IOUtils.copy(inputStream, outputStream);
}
```

EdwardRaff:
Something like that would be better, as it avoids treating potentially arbitrary data as strings and avoids unnecessarily large allocations for larger streams

EdwardRaff:
I've updated the code to work with #240, which does simplify the gzip support. 

For some reason github thinks I changed the whole file... I don't know what that is about.

PR #238, opened 2015-02-04
added awaitInitialization method on SparkBase

### LINKED ISSUE #226: How to check that spark has started without polling?
I have a program that needs to load spark a 'get' resource, but only after spark has started

so my current main is:
````{
spark.start();
loadSparkURL();
}````

Yet oftentimes, since I suppose spark uses an asynchronous start, the URL needs to be refreshed. Is there any way to check without polling that URL every so often?

amarseillan:
To be able to wait for the jetty server to be up before making requests. Fixes #226

### DIFF
--- a/src/main/java/spark/SparkBase.java
+++ b/src/main/java/spark/SparkBase.java
@@ -1,5 +1,7 @@
 package spark;
 
+import java.util.concurrent.CountDownLatch;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


perwendel:
Nice!

amarseillan:
hey @perwendel, actually in my fork of spark I also made the init() method public and it's not called from every route declaration. Also made the route list immutable to end all thread safety problems

Interested in a PR on this?

(this PR is still cool for tests and calling the server from the same code)

PR #239, opened 2015-02-05
change byte[]-string conversion method to a simpler one

amarseillan:


src/main/java/spark/Request.java
```diff
@@ -240,7 +240,7 @@ public String body() {
     private void readBody() {
 		try {
 			bodyAsBytes = IOUtils.toByteArray(servletRequest.getInputStream());
-			body = IOUtils.toString(new ByteArrayInputStream(bodyAsBytes));
+			body = new String(bodyAsBytes);
```
jkwatson: both of these bother me, since we're relying on the default encoding for the system.

amarseillan: I agree, I didn't want to change functionality just yet, just fixing random stuff

perwendel:
I'm sorry, but what is the gain of merging this PR?

amarseillan:
Having 1 conversion instead of 3?
If that aint reason enough then just don't, I just think it's better code
quality

perwendel:
I missed that. Nice tone btw!

amarseillan:
Lol my bad :hand:

perwendel:
:+1:

PR #240, opened 2015-02-09
Adds Serializer class for serializing output data

lordofthejars:
Adds Serializer interface for serializing output data to different formats instead of calling toString.

Before this PR the output was written to outputstream as a byte[] of a call to `toString` method. This works for almost all cases but if you want to return binary content (for example a dynamic generated pdf), then instead of writing the content of the pdf to output, it was written the result of calling `toString` method of a byte[] object which is totally wrong.

For this reason a Serializer abstract class has been created to implement a chain of responsibility pattern. Currently instead of calling `toString` method, Spark tries to resolve the output element. If it is a byte[] or ByteBuffer the output is written to output as is. If it is an InputStream then the inputstream is read and sent as byte[], and the default serializer is doing exactly as it was doing Spark, that is calling toString method.

If the proposed solution is accepted and merged, we can create a new issue which allows users to register its own serializers.

Note that developers are in charge of specifying the content type.

### DIFF
--- a/src/main/java/spark/ResponseTransformerRouteImpl.java
+++ b/src/main/java/spark/ResponseTransformerRouteImpl.java
@@ -38,7 +38,7 @@ public abstract class ResponseTransformerRouteImpl extends RouteImpl {
                                                       ResponseTransformer transformer) {
         return new ResponseTransformerRouteImpl(path, acceptType) {
             @Override
-            public String render(Object model) throws Exception {
+            public Object render(Object model) throws Exception {
                 return transformer.render(model);
             }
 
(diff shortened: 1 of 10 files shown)


vbauer:
:+1:

perwendel:
Thanks for contributing. I'll start reviewing this PR now.

lordofthejars:
Thanks :D I need it for Arquillian Cube hehehe ;)

El dj., 12 març, 2015 a les 15:59, Per Wendel (<notifications@github.com>)
va escriure:

perwendel:
Nice!

PR #244, opened 2015-02-20
Add OSGi bundle configuration to pom so that Spark will build as a bundle.

steve-taylor:
This supersedes pull request #121 (mine) which was conflicted. I resolved the conflict, but overly polluted the history, so here's a clean pull request.

This also supersedes pull request #164 (not mine) which had exactly the same changes and is now also conflicted.

As discussed previously, the only effect of this change is to add OSGi metadata into MANIFEST.MF, allowing the Spark jar to be seen as a bundle by any OSGi container. This change has absolutely no effect on anything other than the generated MANIFEST.MF file.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
     </parent>
     <groupId>com.sparkjava</groupId>
     <artifactId>spark-core</artifactId>
-    <packaging>jar</packaging>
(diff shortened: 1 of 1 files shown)


ooxi:
:+1:

PR #245, opened 2015-02-24
Refactoring in findTargetWithGivenAcceptType 

marcus-sousa:
Hello, 
 I made a refactoring in findTargetWithGivenAcceptType method. All tests are passing, can validate please? Thank U.

Tests run: 69, Failures: 0, Errors: 0, Skipped: 0
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------

### DIFF
--- a/src/main/java/spark/route/SimpleRouteMatcher.java
+++ b/src/main/java/spark/route/SimpleRouteMatcher.java
@@ -164,24 +164,19 @@ public class SimpleRouteMatcher {
         return matchSet;
     }
 
-    // TODO: I believe this feature has impacted performance. Optimization?
+    // TODO: I believe this feature has impacted performance. Optimization? Optimization?
     private RouteEntry findTargetWithGivenAcceptType(List<RouteEntry> routeMatches, String acceptType) {
-        if (acceptType != null && routeMatches.size() > 0) {
-            Map<String, RouteEntry> acceptedMimeTypes = getAcceptedMimeTypes(routeMatches);
-            String bestMatch = MimeParse.bestMatch(acceptedMimeTypes.keySet(), acceptType);
 
-            if (routeWithGivenAcceptType(bestMatch)) {
-                return acceptedMimeTypes.get(bestMatch);
-            } else {
-                return null;
-            }
-        } else {
-            if (routeMatches.size() > 0) {
-                return routeMatches.get(0);
-            }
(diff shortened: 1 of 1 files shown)


perwendel:
I'm trying to validate. A question:
routeWithGivenAcceptType is not called anymore. Isn't that needed?

solivavinicius:
Hi, 
 I do not need **routeWithGivenAcceptType**.

**MimeParse.bestMatch(acceptedMimeTypes.keySet(),acceptType);** returns NO_MIME_TYPE if it fails and **routeWithGivenAcceptType** only verifies that the return of this operation is not equal to NO_MIME_TYPE to perform an operation on the map, I believe that this operation can be performed without any check as if the key does not exist on the map return the operation will be null.

Correct ?

src/main/java/spark/route/SimpleRouteMatcher.java
```diff
@@ -164,24 +164,19 @@ private boolean routeWithGivenAcceptType(String bestMatch) {
         return matchSet;
     }
 
-    // TODO: I believe this feature has impacted performance. Optimization?
+    // TODO: I believe this feature has impacted performance. Optimization? Optimization?
     private RouteEntry findTargetWithGivenAcceptType(List<RouteEntry> routeMatches, String acceptType) {
-        if (acceptType != null && routeMatches.size() > 0) {
+
+        if (routeMatches.isEmpty()) {
+            return null;
```
dinesh707: can "routeMatches" can be null? if spark has apache commons use CollectionUtils.isEmpty() else, just check list is null too

src/main/java/spark/route/SimpleRouteMatcher.java
```diff
@@ -164,24 +164,19 @@ private boolean routeWithGivenAcceptType(String bestMatch) {
...
-            }
-        } else {
-            if (routeMatches.size() > 0) {
-                return routeMatches.get(0);
-            }
+            return acceptedMimeTypes.get(bestMatch);
         }
-
-        return null;
+        return routeMatches.get(0);
```
dinesh707: is this a good change ? whats the basis for selecting the first one ?. if its a problem of caller then I would throw an exception

tipsy:
Class has been deprecated.

PR #250, opened 2015-03-10
don't include logging in distribution

benjyz:
Including slf4j at runtime is not needed. It can override logging downstream, and should be excluded.


benjyz:
@perwendel would you mind merging this?

laktak:
+1

Issue #251, opened 2015-03-11
Session Invalidation Error


melvinma:
Session invalidate and then recreated failed for me - The reason is, after invalidate a session, the session and the invalid HttpSession was not cleared out. I recommend the following code change in spark: in Request class, whenever access session variable, check if session is valid, if not valid, create a new session variable.

---- code having issues ----
			Session session = request.session(false);
			if (session!= null) session.invalidate();			
			session = request.session(true);
			session.attribute(SessionKeys.USER_ACCESS_TOKEN, "1234567");

---- exception I got ----

java.lang.IllegalStateException: null
	at org.eclipse.jetty.server.session.AbstractSession.checkValid(AbstractSession.java:104) ~[jetty-server-9.0.2.v20130417.jar:9.2.7.v20150116]
	at org.eclipse.jetty.server.session.HashedSession.checkValid(HashedSession.java:79) ~[jetty-server-9.0.2.v20130417.jar:9.2.7.v20150116]
	at org.eclipse.jetty.server.session.AbstractSession.changeAttribute(AbstractSession.java:509) ~[jetty-server-9.0.2.v20130417.jar:9.2.7.v20150116]
	at org.eclipse.jetty.server.session.AbstractSession.setAttribute(AbstractSession.java:461) ~[jetty-server-9.0.2.v20130417.jar:9.2.7.v20150116]
	at spark.Session.attribute(Session.java:55) ~[spark-core-2.1.jar:na]

perwendel:
Sorry for not looking at this earlier. I'm not the author of the session functionality so this might be a stupid question. If the underlying session is invalidated, shouldn't attribute() throw IllegalStateException?

perwendel:
I looked a bit more. I guess it should be possible to create a new session after the old has been invalidated?

jazdw:
There is no way to create a new session after the previous session has been invalidated. I had to work around this by using the underlying "raw" session. We need to be able to do this in order to change the session ID when logging in.

jazdw:
Have a look at the Jetty Request.getSession(boolean) method. It checks if the session is valid and creates a new one if not.

tipsy:
Fixed in 2.5

PR #252, opened 2015-03-11
issue #251:Session Invalidation Error

melvinma:
This is a fix for #251 . When Session is invalidate, the HttpSession variable is set to null. In the Request object, when using session, always check session.raw(). 

This is may not be a cleanest solution. Alternatively, Session could reference back to Request, so when invalidate, it cleans out the session variable in the Request object.

### DIFF
--- a/src/main/java/spark/Request.java
+++ b/src/main/java/spark/Request.java
@@ -372,7 +372,7 @@ public class Request {
      * @return the session associated with this request
      */
     public Session session() {
-        if (session == null) {
+        if (session == null || session.raw() == null) {
             session = new Session(servletRequest.getSession());
         }
         return session;
@@ -388,7 +388,7 @@ public class Request {
      * <code>create</code> is <code>false</code> and the request has no valid session
      */
     public Session session(boolean create) {
-        if (session == null) {
+        if (session == null || session.raw() == null) {
             HttpSession httpSession = servletRequest.getSession(create);
             if (httpSession != null) {
                 session = new Session(httpSession);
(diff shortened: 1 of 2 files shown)


perwendel:
With this fix calling methods on the invalidated session object will throw nullpointers instead of IllegalStateExceptions which doesn't seem right. I guess it should be flagged in another way that the session has been invalidated!?

perwendel:
ping @melvinma see previous comment pls.

src/main/java/spark/Request.java
```diff
@@ -372,7 +372,7 @@ private void initQueryMap() {
      * @return the session associated with this request
      */
     public Session session() {
-        if (session == null) {
+        if (session == null || session.raw() == null) {
```
dinesh707: update java doc.

src/main/java/spark/Request.java
```diff
@@ -388,7 +388,7 @@ public Session session() {
      * <code>create</code> is <code>false</code> and the request has no valid session
      */
     public Session session(boolean create) {
-        if (session == null) {
+        if (session == null || session.raw() == null) {
```
dinesh707: this "session == null || session.raw() == null" part is duplicated. And I would suggest another simple method which explains what it means (eg : isSessionRawNull() )

src/main/java/spark/Session.java
```diff
@@ -113,6 +113,7 @@ public void maxInactiveInterval(int interval) {
      */
     public void invalidate() {
         session.invalidate();
+	session = null;
```
dinesh707: update doc

tipsy:
This was fixed in a different PR.

PR #253, opened 2015-03-17
Reduced some boilerplate by adding SimpleRoute and associated methods.

martylamb:
SimpleRoute can be used in place of Route for the simpler route-adding
methods.  SimpleRoute provides a no-arg handle() method so that routes
can be added like this:

    get("/path", () -> "Hello");

instead of:

    get("/path", (request, response) -> "Hello");

Users of SimpleRoute can still access the request and response via the
new static methods request() and response() in SparkBase.  Internally,
these are stored within ThreadLocals managed by MatcherFilter.

At the moment, only the METHOD(String, SimpleRoute) methods have been
added.  If this path looks fruitful I'll implement the rest.

### DIFF
--- a/src/main/java/spark/SimpleRoute.java
+++ b/src/main/java/spark/SimpleRoute.java
@@ -1,0 +1,13 @@
+package spark;
+
+/**
+ * Created by Marty Lamb on 2015-03-17.
+ */
+public interface SimpleRoute {
+
+    /**
+     * Invoked when a request is made on this route's corresponding path e.g. '/hello'
+     */
+    Object handle()  throws Exception;
+
+}
--- a/src/main/java/spark/Spark.java
+++ b/src/main/java/spark/Spark.java
@@ -40,8 +40,7 @@ public final class Spark extends SparkBase {
     // Hide constructor
     private Spark() {
     }
-
-
+    
     /**
      * Map the route for HTTP GET requests
      *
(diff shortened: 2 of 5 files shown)


chillenious:
Imho it is great that Spark is pretty minimal right now. The project I work on that uses Spark has a bunch of specialized Routes, and they are easy enough to add yourself (possibly together with filters, e.g. to set up and tear down thread locals). Adding more of these to the core project will only increase the conceptual surface of the framework. If anything, i think they'd go better into examples of such. My 2c.

martylamb:
It's definitely easy to add for oneself - that's what I've been doing for a while, in fact.  But I've found the pattern useful often enough that I think it's a sensible addition to the core.

One benefit that's not obvious from my commit comment is that not only do you not have to have your Route accept the Request and Response as parameters (admittedly not that big of a deal), but you're also relieved from having your Route pass them around to any other methods/objects it uses to handle the request.  Any of those can simply call request() or response() in the same way.

I do get the concerns about complexity and that's always something to watch.  But in this case I would argue that the net result is simplification.

Now if only travis-ci would provide me some build output so I could see why there's a build failure I can't reproduce locally.

martylamb:
Aha.  Build failure due to refused connections in tests.  Looking through travis build history, this just... seems to happen from time to time.  I do not believe the build failure has anything to do with the changes in the pull request.

pkafel:
@martylamb Got the same problem. If you will figure out what to do with it I hope you will share.

perwendel:
Build retriggered.

perwendel:
I'm skeptical to adding this since one thing I try to enforce (for consistency) is just provide one way of doing things.

However, before deciding I would like to get more input from the community to try to get a feeling what the opinion is about this.

martylamb:
Sure, thanks for taking a look.

tipsy:
Won't add this to Spark core. We'll create a new project for stuff like this ("Spark sugar") soon.

Issue #193, opened 2014-09-09
Why did you use a Filter instead of a Servlet?


MikeMitterer:
I'm using Shiro for Access-Verification. Shiro is a Filter which makes sense but if Spark is also a Filter it always depends on "who is first" in the filter chain if it works or not.
I took your Filter-code change the Servlet-according-things and not it works nicely together with Shiro.
So again - is there a reason for the SparkFilter?

Here is my Servlet:

``` Java
package at.mikemitterer.webapp.guice.spark;

import com.google.inject.Singleton;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import spark.route.RouteMatcherFactory;
import spark.servlet.SparkApplication;
import spark.webserver.MatcherFilter;

import javax.inject.Inject;
import javax.servlet.ServletConfig;
import javax.servlet.ServletException;
import javax.servlet.http.HttpServlet;
import javax.servlet.http.HttpServletRequest;
import javax.servlet.http.HttpServletRequestWrapper;
import javax.servlet.http.HttpServletResponse;
import java.io.IOException;

/**
 * @since 09.09.14, 12:14
 */
@Singleton
public class SparkServlet extends HttpServlet {
    @SuppressWarnings("unused")
    private static final Logger logger = LoggerFactory.getLogger(SparkServlet.class);

    private static final String SLASH_WILDCARD       = "/*";
    private static final String SLASH                = "/";
    private static final String FILTER_MAPPING_PARAM = "filterMappingUrlPattern";

    private String filterPath;
    private MatcherFilter matcherFilter;

    final SparkApplication application;

    @Inject
    public SparkServlet(final SparkApplication application) {
        this.application = application;
    }

    @Override
    public void init(final ServletConfig config) throws ServletException {
        super.init(config);
        Access.runFromServlet();

        application.init();

        filterPath = getConfigPath(config);
        matcherFilter = new MatcherFilter(RouteMatcherFactory.get(), true, false);
    }

    @Override
    protected void service(final HttpServletRequest req, final HttpServletResponse resp) throws ServletException, IOException {
        //super.service(req, resp);

        HttpServletRequest httpRequest = (HttpServletRequest) req; // NOSONAR

        final String relativePath = getRelativePath(httpRequest, filterPath);

        logger.debug(relativePath);

        HttpServletRequestWrapper requestWrapper = new HttpServletRequestWrapper(httpRequest) {
            @Override
            public String getRequestURI() {
                return relativePath;
            }
        };
[...]

PR #254, opened 2015-03-17
Add servlet instead of filter. (Fix Issue #193.)

avarabyeu:
Add servlet instead of filter. Keep old filter implementation for backward compatibility. Business logic is not changed. Introduced MatcherHandler (which is old MatcherFilter, but without FilterChain), SparkServlet and SparkHandler (just for keep external resources configuration the same for SparkFilter and SparkServlet)


src/main/java/spark/servlet/SparkHandler.java
```diff
@@ -0,0 +1,121 @@
...
+
+    static boolean staticResourcesSet = false;
+    static boolean externalStaticResourcesSet = false;
+
+    /**
+     * Configures location for static resources
+     *
+     * @param folder the location
+     */
+    public static void configureStaticResources(String folder) {
```
dinesh707: can you make those 2 methods to share code?. both has quite similar code

tipsy:
#671

Issue #266, opened 2015-04-11
parameters are returned undecoded


pvg:
Request.params returns undecoded strings which makes any parameters with, say, a space or non-latin character inconvenient to use. Even a low-level API like HttpServletRequest.getParameter does this for you. 

One (possibly naive) solution is to tweak line 90 in MatcherFilter.java from

```
    String uri = httpRequest.getRequestUri(); // NOSONAR to

    String uri = httpRequest.getPathInfo(); // NOSONAR
```

This works and passes the embedded Jetty tests. Inexplicably, it fails all the ServletTest tests because in that setup, getPathInfo returns null. Maybe this is something that can be fixed by tweaking the config of the web app example? It seems really odd that the 'embedding in an arbitrary servlet container' example/test fails on the failure of a standard servlet API call. 

Of course, one could also make MatcherInfo decode the getRequestURI although getPathInfo looks like the more appropriate call for route matching. 

Edit: I'd completely missed the SparkFilter wrapper which fakes out the getRequestUri for the servlet container integration case. I went ahead and changed it to override getPathInfo and made a small change to FilterTools to do the urldecoding and submitted a PR for discussion 

https://github.com/perwendel/spark/pull/267

tipsy:
Can this be closed now?

pvg:
No, because the implementation is wrong. It's just less wrong than it used to be. The problem is encoding/decoding standards are slightly different for different parts of a URI and the current implementation uses URIDecoder which is intended for query strings, not paths. Unfortunately, there is no built-in API for handling this, particularly in the servlet case where there are no guaranteed dependencies like jetty or commons http-client. The solutions is, perhaps, to lift the code just for this from an open source servlet implementation or possibly from guava, without introducing another fat dependency.

tipsy:
Not sure if fixed. Closing this because we need to clean up. @pvg Please create a new issue if this is still a problem.

PR #267, opened 2015-04-11
Make params/routes unicode strings

pvg:
A tentative PR addressing issue https://github.com/perwendel/spark/issues/266


pvg:
There's something incomplete/not quite right with this still, if you change BooksIntegrationTest line 30 to

    private static String AUTHOR = "Лев Толстой";

the test fails. This should be passing (and, in fact, should be part of the test). I'll take a closer look at it in the next couple of days and send you another PR.

pvg:
I was looking at the wrong test, added a test in the right place, PR #276

PR #270, opened 2015-05-07
halt methods return HaltException instead of void

amanteaux:
Halt methods should return HaltException instead of void to enable writing ```throw halt(401)```.
Currently in a route, one has to write ```halt(401); return null;```. The part ```return null;``` is useless and confusing since it will never be executed.


amarseillan:
:+1:

guicamest:
:+1:

PR #271, opened 2015-05-14
Response cookie removal setSecure

Heedster:


perwendel:
Please add description. Why is this needed when removing the cookie?

PR #276, opened 2015-05-18
tweak to include roundtrip test for correct encoding of path-based sp…

pvg:
…ark params


perwendel:

pvg:
If you're going to think about this, might be better to just take another look at #267 because the test thing is kind of secondary (and I think it might be better to overhaul that test harness in general, which I have a first pass of sitting about here somewhere). The bigger issue is, what's a good way to fix #267? Is it ok to introduce another dependency? Should some code be lifted from a bigger library to avoid that?

perwendel:
:+1: 
"Copy paste" merge due to branch conflicts.

PR #279, opened 2015-05-27
Use official HTTP 401 Unauthorized message name

lukaseder:
I personally appreciate the kicks and giggles that arise from the odd usage of this language, but it feels a bit weird to have this kind of message in what would potentially be production code.

In other words ... Rewording this can only be in the best of your project's interests ;)


perwendel:
Yeah, this class shouldn't be there at all. I guess it must have gotten in with the lambda re-write last year. It's removed now though. Fuck off LambdaTest!

PR #283, opened 2015-05-29
Add queryParamsValues to Request

serandel:
This method supports URLs like:

    http://host/resource?id=foo&id=bar

returning `request.getParamsValues("id")`as `["foo", "bar"]`

It simply delegates to the appropiate method in `HttpServletRequest`. Without it, the non-intuitive and a bit cumbersome way of getting this result is:

    QueryParamsMap paramMap = request.queryMap("id");
    if (paramMap.hasValue()) {
        whatever(paramMap.values())
    }

### DIFF
--- a/src/main/java/spark/Request.java
+++ b/src/main/java/spark/Request.java
@@ -263,6 +263,17 @@ public class Request {
     public String queryParams(String queryParam) {
         return servletRequest.getParameter(queryParam);
     } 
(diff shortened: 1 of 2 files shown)


perwendel:

perwendel:
:+1:

serandel:
:)

PR #285, opened 2015-05-31
method `handle` of TemplateViewRoute can throw Exceptions

bryce-anderson:
This restores the symmetry between the template operators and the other route types.


perwendel:

PR #305, opened 2015-06-23
Automatically decode query string values

alediaferia:
As per the current doc in QueryParamsMap I would expect QueryParamsMap values to be URL-Decoded but this isn't actually happening. A test to demonstrate this has been added and a patch is here to resolve the issue.


alediaferia:
Nevermind. Not a Spark fault.

PR #308, opened 2015-07-14
Instantiated API for Spark

ggalmazor:
Hi again!

#167 was out of date and I've created a new fork from the latest base code with the same changes. The goal here is to be able to use Spark with its default static API **and** by instantiating a SparkAPI instance.


src/main/java/spark/Spark.java
```diff
@@ -36,11 +34,19 @@
  *
  * @author Per Wendel
  */
-public final class Spark extends SparkBase {
+public final class Spark {
```
jkwatson: I'd prefer if this weren't final. Is there a good reason why it should be?

src/main/java/spark/SparkAPI.java
```diff
@@ -0,0 +1,1132 @@
...
+
+    /**
+     * Set the port that Spark should listen on. If not called the default port
+     * is 4567. This has to be called before any route mapping is done.
+     * If provided port = 0 then the an arbitrary available port will be used.
+     *
+     * @param port The port number
+     */
+    public synchronized void port(int port) {
+        if (initialized) {
```
jkwatson: To my sensibilities, it would be preferable to separate the pieces that have to be initialized before the routes, and the routes themselves. Then, all of this extra complication could be eliminated in favor of an immutable core configuration (the pre-route bits), and and immutable set of filters/routes.  This would remove the necessity of managing the latch, and all the state.

jkwatson:
This is definitely a move in the right direction, but I wonder if it might be better to do some real cleanup along with this, and release it as spark 3.0.  In particular, it would be really great to separate the pre-route pieces from the route-based pieces, and make both of the pieces immutable, so that the complicated synchronization/latch management could go away altogether.

heuermh:
+1

I also agree with the suggestion from @jkwatson about separate immutable pre-route and route-based pieces.

src/main/java/spark/Spark.java
ggalmazor: @jkwatson originally it was already final and I've tried to limit my changes to a bare minimum. I guess that after this PR is merged we could change if it gets traction.

src/main/java/spark/SparkAPI.java
ggalmazor: Totally agree with you on this one. I'd wait first to merge this PR though.

src/main/java/spark/Spark.java
jkwatson: fair enough.

perwendel:
Tx, will have a look at this after my vacation (the week after next)

benhyland:
This is a very timely PR. We are currently considering moving away from Spark due to the overhead and coupling introduced by maintaining our own fork for a non-static interface. This would be a shame as other than that Spark fits our needs very well.
The non-static interface is more in line with our general practice, but the killer requirement is to be able to run multiple independent Spark-driven webapps in a single external Resin instance (i.e. a single JVM) without them trampling one another.
Please consider merging this sooner rather than later!

perwendel:
@benhyland I'll look at this ASAP. The pros and cons of an Instance API of Spark has been discussed a lot before and I'm still unsure of what is the best way to get this in Spark. I need to figure out if this is the best way or a more "sparkish" code style of running multiple web-apps is possible. The only thing holding back this is that I'm very protective of the Spark programming paradigm (which of course is based on static methods/imports)

benhyland:
@perwendel thanks very much!

perwendel:
@ggalmazor @jkwatson @heuermh @benhyland

I've had a look at the code and the implementation looks good. Why I'm not merging this is since I'm not at all sure that I think an open non-instance API is a good thing. Let me try to explain:

I do know the limitations that a static API creates and I know that a lot of people think static APIs are, almost always, wrong. The thing is, Spark is not designed just to be small and simple. A big reason, perhaps the biggest, was to design the API to be as declarative as possible and not imperative. To achieve and enforce this the API have to be static. To introduce an instance API would be a step in the wrong direction. 

I try not be dogmatic when developing software but when it comes to APIs in general and specifically Spark I'm of the very strong opinion that there should only be ONE way of doing things. Not two ways of achieving the same thing. I do believe this consistency combined with the declarative API is a very big reason to Spark being as succesful as it is and also makes it differ from the other frameworks.

Adding functionality for allowing multiple Spark web apps in a single instance is another thing. This is functionality I think is good and plan to have. It's harder to achieve a good way of doing this in a static API but this I am working on. I will probably use parts of this PR to achieve this.

borud:
I'm not sure I understand how the static API makes for a less imperative style since it pretty much excludes any functional style of programming given that one, in essence, ends up manipulating global, mutable state.

The biggest problem we have on my project right now is that we can't do proper unit testing unless we hobble the unit testing system severely by firing up a new JVM for each test to guarantee sufficient predictability and separation.  That's the price of global state.  And that's extremely slow and it imposes lots of problems down the line if someone were to re-use our code as part of another project.

perwendel:
@borud 
The fact that a global state is altered under the hood does not, at lest to me,  make the API imperative. When routes are defined the code is clearly declarative.
I don't know if we have the same understanding about what's declarative/imperative and what's not but I'll try to explain how I see it:

```java
get("/hello", (req, res) -> "Hello World");
```

is a lot more declarative than:

```java
SparkAPI spark = new SparkAPI();
spark.get("/hello", (req, res) -> "Hello World");
```

The first snippet just describes what the user wants, add a get route for /hello, hence declarative. The second snippet introduces the 'spark' instance which has no declarative function at all and introduces the imperative part that describes the SparkAPI being instantiated and that instance is later used for adding routes.

I do understand that a static API poses some difficulties regarding unit testing, but that is the cost that has to be taken for keeping a clean and declarative API. 

If I want unit tests for my business logic I would create "handlers", extending Route, that holds the business logic and pass them as parameter to my route definition, e.g:

```java
post("/books", new BooksCreateHandler());
```

Those handlers should be easy to unit test. There's also a tutorial on Spark and unit testing: https://sparktutorials.github.io/2015/07/30/spark-testing-unit.html

ribomation:
Per, your overall line of argument is as people have suggested to replace the static api for an instance ditto, when in fact we want to have both. Just wrap the static api around a singleton and it will be unaffected, keeping the nice and easy api we all like. 

However, it is when we have written all kickoff spark apps and start using it for serious applications, demands for a more elaborated api arises. Such as being able to extend or override/replace certain parts, launch several servers in the same jvm listening on different ports and more.

perwendel:
@ribomation then you have missed what I wrote in a previous comment: "I try not be dogmatic when developing software but when it comes to APIs in general and specifically Spark I'm of the very strong opinion that there should only be ONE way of doing things."

I don't know how to address the extension/override part since I don't know what parts you mean you'd like to override replace? If it's your business logic, the example in my last post should be sufficient (BooksCreateHandler). If it's extending/overriding Spark itself I would not use it that way (if there was an instance API, I would probably make the class final to enforce consistency)

As I said before, launch several servers is a functionality that I plan to integrate in the future, however not via an instance API.

This is what it boils down to. When I created Spark, I basically put all the time into the API/programming paradigm. What I was trying to create was a paradigm that I saw was missing in java. I had to break tons of java conventions and recommendations to achieve what I was aiming for. I have since then been extremely protective of the API because I'm fairly sure that it's the biggest attraction of Spark, not that it's small etc. The whole idea with Spark was creating a declarative API with a strict paradigm on doing things, and just have one way of doing this. To introduce an alternative would be to go against the core idea of Spark. The idea I believe is the single biggest reason to Spark being a success.

I might sound stubborn about this, and in a sense I am. It's just that I think it would be bad for Spark in the long run to introduce an instance API.

borud:
@perwendel We are talking about slightly different things I think.  You are talking about syntax while I'm talking about semantics.  The more compact syntax of the static API "looks" more declarative, but when executed it mutates global state which makes it tricky to say much about the state of the system at any given time while mutating it without resorting to external coordination.  It also comes with all the problems of global state.  This isn't unlike programming with global variables in any programming language, with all the problems this entails.

In a truly declarative system there are no intermediary states before the object is exposed.  The declaration is parsed and the object it describes is instantiated in its final form before being exposed to the calling code.  The programming equivalent to a declarative description would be an immutable object.

I think the lengths one has to go to in unit tests is a clear illustration of how awkward messing with global, mutable state is.  And I don't see a solution for performing proper tests that verify that the routing is correct.  (For instance let's say you have an important `before()` and you want tests in several locations that your application will behave correctly if the `before()` is accidentally dropped -- even that becomes tricky since you will have to be very careful about the sequencing and concurrency of those tests -- meaning they will be slower, more error prone, harder to implement and much harder for newcomers to understand)

**Having an instance based core API does not mean you have to sacrifice your preferred API.   It just means that you can have both.  You can make a wrapper class that contains a singleton Spark instance, and if you call the wrapper class Spark you should be able to run all existing code without any changes.**

And for an instance-based API, you will essentially only have 2-3 more lines of code.  That will more than make up for having to write many orders of magnitude more lines of unit test code, loss of performance and increased complexity:

    Spark spark = Spark.newBuilder()
        .get("/hello", (req, res) -> "Hello World")
        .post("/somewhere"), (req, res) -> "Got post")
        .port(8080)
        .build()
        .start();

If, in the above example, you imagine that `Spark` instances are immutable and that you provide a `SparkBuilder` class to help you build a `Spark` instance, you get much closer to the semantics of a declarative system in that no intermediate states are exposed.  The Spark instance either exists or it doesn't.  It is also clearly defined where the instance comes into existence and where you start it.  You also have the option of ensuring that inconsistent instances are not allowed to be instantiated.

perwendel:
@borud 
I agree that we're talking about two different things. I am 100% focusing on the API, Not the implementation. I still hold the position that the static API is the most declarative API. I will soon change the underlying implementation to work with instances for different reasons. 

However, as I said in previous comments, the API has highest priority. Being as easy to unit test as possible has not. Having two APIs for achieving the same thing isn't an option, I've explained in previous comments why.

The syntax:

```java
Spark spark = Spark.newBuilder()
    .get("/hello", (req, res) -> "Hello World")
    .post("/somewhere"), (req, res) -> "Got post")
    .port(8080)
    .build()
    .start();
```

does not attract me, at least if I am looking for a Sinatra like framework. A fluent interface has its perks and there are definitely areas when it's preferable, but when it comes to adding routes, I prefer the static approach anyday.

One thing I haven't explained before is that I see Spark as part of a movement to "rejuvenate" java and free it from certain dogmas I believe is holding it back. The view that "static is bad" is one of those dogmas. Static can be very bad, but when it comes to designing declarative APIs (just the APIs) I think it's perfect (combined with static imports).

To be extra clear, I do believe the current limitation of one spark server per JVM instance is a problem and is the next thing I'll work on.

i386:
@perwendel looks like there is enough interest from the community (myself included) to move to a instance based API for spark.  I believe there have been a few attempts in this direction and contributing their time but their efforts haven't really lined up to what you think is best for the project. 

Would it be possible for you to layout what your plans are somewhere (Google doc, wiki, etc) and have the community contribute their thoughts? It would also be great to capture some sort of immutable philosophy about the project (for example, the API should come first, keeping the framework micro, etc)

AleksanderGrzybowski:
I would be really happy to use instantiated API in some form. Spark is a really elegant solution to a lot of my needs, I used it a few times, but testing is very cumbersome, and polluting global state in unpredictable way is a big no-no for me.

perwendel:
@i386 
There is interest from parts of the community. I think it's a smaller part than this "thread" makes it look like. I've also explained it a couple of times why an instance based API is not currently an option. The idea of laying out the plans is very good. Even better is the idea of explaining the immutable philosophy behind the project. I've been meaning to have the explanation for the static API somewhere but capturing the philosophy would be even better. Thanks!

perwendel:
@AleksanderGrzybowski 
As shown in both Spark tutorials and other forums testing is not very cumbersome, it's more a matter of how you design you code.

"polluting global state in unpredictable way" kind of sums up what I reject with a lot of the arguments for an instance API. If coded correctly, as Spark is, this is not a problem, at least not from my perspective.

IMHO there is a irrational fear of `static` in the java world that arose with the advent of DI, test with mocks, etc. As I see it, those design rules/conventions holds Java back. 

Static (where fit) + lambdas + method references + breaking conventions combined can make modern , functional and really cool Java APIs.

As per my last comment, I will try to summarize and capture the "immutable" philosophy behind Spark so people can read and understand, agree or disagree, the decisions for the project.

i386:
@perwendel that's great stuff - I look forward to your write ups! 👏

ggalmazor:
Well... I can't say that I agree with your decision here.

The reality is that this PR (and the previous one) didn't get much traction from the community (just 5-10 people interested on going this way) but I'm disappointed that after working on this and trying to make a better Spark, the decision is made unilaterally claiming some hard to believe claims about rejuvenating the JVM.

I use Spark personally and professionally since before version 1.0 and this PR comes from my experience designing complex apps. Spark is very difficult if you're designing uncoupled software (with hexagonal architecture, in my case) and I think that this should be Spark's concern if it has the vocation of becoming a major player on Java's modern stack of frameworks and libs.

mread:
Agree, very disappointed and won't be introducing Spark anywhere new as a
result.

ggalmazor:
Hi @mread. I wrote that comment before realizing that some of this PR has been merged in #344 

I think that this is a step in the right direction (still disappointed, though).

benhyland:
@perwendel thanks for explaining your approach further. I certainly appreciate your points about the deleterious effect on program design that mocks, IoC and so on have had in the past.
I want to point out that a lot of design patterns in the Java ecosystem originate from a time well before reasonable functional approaches were sufficiently succinct, and in many cases also from before parametric polymorphism entered the language. We shouldn't judge folks too harshly for using what they had available at the time :)
I am still not really seeing an advantage to shared mutable state in this instance but you have clearly made a decision, and that in itself is valuable for everyone concerned with this issue.
Can you provide similar clarity on the problem I mentioned before? Do you have any more info on whether or when Spark might be able to support multiple webapps defined in separate compilation units being run safely in the same JVM, in the multithreaded environment of a servlet container? You mentioned that you plan to support several servers but without more detail I'm not sure we are talking about the same thing. Thanks.

Issue #196, opened 2014-09-19
Adding simple websocket feature


mustafaakin:
I am planning to extend spark for my own usage, and if it is good enough, share it with the community. I really need web sockets, it is essential. I do not have much experience with Netty, but I need to understand some concepts. 

```class SparkServerImpl implements SparkServer {``` takes a Handler object. 

It seems that websockets in jetty is just used like in this page: https://github.com/jetty-project/embedded-jetty-websocket-examples/blob/master/native-jetty-websocket-example/src/main/java/org/eclipse/jetty/demo/EventServer.java 

It is just another handler:

```
        ServletContextHandler context = new ServletContextHandler(ServletContextHandler.SESSIONS);
        context.setContextPath("/");
        server.setHandler(context);
        
        // Add a websocket to a specific path spec
        ServletHolder holderEvents = new ServletHolder("ws-events", EventServlet.class);
        context.addServlet(holderEvents, "/events/*")
```

So is it okay that I websocket its handler to following part in the `SparkServerImpl`:

```
           List<Handler> handlersInList = new ArrayList<Handler>();
            handlersInList.add(myWebSocketHandler); <-----
            handlersInList.add(handler);
```

**Comment by fge** [comment]
Hello,

I am very, very deeply interested in this feature; what is its current status?

**Comment by tchoulihan** [comment]
Same here, would love this feature.

**Comment by davidvollmar** [comment]
Getting a basic proof of concept working:

Add the jetty websocket-server to your `pom.xml`
```xml
        <dependency>
            <groupId>org.eclipse.jetty.websocket</groupId>
            <artifactId>websocket-server</artifactId>
            <version>${jetty.version}</version>
        </dependency>
```

And use the sample from jetty provided. 

How about exposing a way to add your own handlers without needing to hack it in the source?

fge:
@davidvollmar that still doesn't tell how to use it with spark though... I'm in the process of trying to hack a simple example using jetty and jetty alone since from this pom.xml alone I am unable to see how to exploit it with spark.

Tips welcome...

davidvollmar:
Hi Francis,

See https://gist.github.com/davidvollmar/d3ff37be35f218dbdadc for the files you need. You'll need to replace `SparkBase.java` by the given one or add 
```java
 // expose setExternalHandlers, not available normally.
public static void setExternalHandlers(List<Handler> handlers){
server.setExternalHandlers(handlers);
}
``` 
to your `SparkBase.java`

[...]

Issue #65, opened 2013-04-30
Support for WebSocket


panterose:
It's available in Jetty 9, that would be a nice feature.

yrro:
Dupe of #169?

Gurey:
+1

PR #309, opened 2015-07-17
Added basic WebSocket support in the embedded server

nacx:
This PR is based on the comments in https://github.com/perwendel/spark/issues/196.
Adds basic support for WebSockets in the embedded server.

### DIFF
--- a/src/main/java/spark/SparkBase.java
+++ b/src/main/java/spark/SparkBase.java
@@ -277,6 +277,12 @@ public abstract class SparkBase {
      * @param timeoutMillis The max idle timeout in milliseconds.
      */
     public static void webSocketIdleTimeoutMillis(int timeoutMillis) {
+	if (initialized) {
+            throwBeforeRouteMappingException();
+        }
+	if (runFromServlet) {
+            throw new IllegalStateException("WebSockets are only supported in the embedded server");
+        }
 	webSocketIdleTimeoutMillis = Optional.of(timeoutMillis);
     }
 
--- a/src/main/java/spark/SparkBase.java
+++ b/src/main/java/spark/SparkBase.java
@@ -1,5 +1,10 @@
 package spark;
 
+import static java.util.Objects.requireNonNull;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 
 import org.slf4j.Logger;
@@ -32,9 +37,12 @@ public abstract class SparkBase {
     protected static String staticFileFolder = null;
     protected static String externalStaticFileFolder = null;
 
+    protected static Map<String, Class<?>> webSocketHandlers = null;
+
     protected static int maxThreads = -1;
     protected static int minThreads = -1;
     protected static int threadIdleTimeoutMillis = -1;
+    protected static Optional<Integer> webSocketIdleTimeoutMillis = Optional.empty();
 
     protected static SparkServer server;
     protected static SimpleRouteMatcher routeMatcher;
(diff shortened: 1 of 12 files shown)


src/main/java/spark/SparkBase.java
```diff
@@ -358,7 +388,7 @@ protected static void addFilter(String httpMethod, FilterImpl filter) {
                                                    + "'", filter.getAcceptType(), filter);
     }
 
-    private static synchronized void init() {
+    public static synchronized void init() {
```
nacx: Note this has been made public to allow calling N times the `webSocket` method, and allow to start Spark serving only WebSockets. If no routes are added, users should be able to manually call `init()` after having declared the websocket handlers.

perwendel:
Tx, will have a look at this after my vacation (the week after next)

joakime:
This can be better handled with use of [`org.eclipse.jetty.websocket.server.WebSocketUpgradeFilter`](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/websocket/server/WebSocketUpgradeFilter.html) in Jetty 9.2.+

Use its [`.addMapping(PathSpec, WebSocketCreator)`](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/websocket/server/WebSocketUpgradeFilter.html#addMapping-org.eclipse.jetty.websocket.server.pathmap.PathSpec-org.eclipse.jetty.websocket.servlet.WebSocketCreator-) method.

perwendel:
I've merged the PR that bumps the jetty version to 9.2.11.v20150529. Do you (@nacx) modify this PR with the tip from @joakime?

Btw, this PR looks good!

nacx:
Sure! I'll rebase the branch and update the PR.

nacx:
@perwendel @joakime I've rebased and squashed the commits into a single one and addressed the comments. Please have a look!
Using the `WebSocketUpgradeFilter` has made possible to get rid of the `WebSocketServlet` so it is no longer exposed to users. They just need to pass the handler class properly annotated without having to extend the servlet.

src/main/java/spark/SparkBase.java
```diff
@@ -241,6 +249,38 @@ public static synchronized void externalStaticFileLocation(String externalFolder
...
+        webSocketHandlers.put(path, handler);
+    }
+
+    /**
+     * Sets the max idle timeout in milliseconds for WebSocket connections.
+     *
+     * @param timeoutMillis The max idle timeout in milliseconds.
+     */
+    public static void webSocketIdleTimeoutMillis(int timeoutMillis) {
+	webSocketIdleTimeoutMillis = Optional.of(timeoutMillis);
```
perwendel: no if (initialized) check and exception if already initialized?

src/main/java/spark/SparkBase.java
```diff
@@ -241,6 +249,38 @@ public static synchronized void externalStaticFileLocation(String externalFolder
...
 
     /**
+     * Maps the given path to the given WebSocket handler.
+     * <p>
+     * This is currently only available in the embedded server mode.
+     *
+     * @param path the WebSocket path.
+     * @param handler the handler class that will manage the WebSocket connection to the given path.
+     */
+    public static synchronized void webSocket(String path, Class<?> handler) {
```
perwendel: are the no restrictions on class type here (essentially in WebSocketCreatorFactory.create(webSocketHandlers.get(path))?

nacx: WebSocket handler classes can be arbitrary classes. Users can:

* Implement the `WebSocketListener` interface (or extend the convenient `WebSocketAdapter`).
* Use an arbitrary class and use the websocket annotations, as seen in [the examples](https://github.com/nacx/spark/blob/ws/src/test/java/spark/examples/websocket/EchoWebSocket.java).

That's why I left it wihtout restrictions, but I'll add a check to require the annotation or the listener interface.

perwendel:
@nacx Looks good. One comment and one question.

nacx:
Will add the checks for the initialized and the websocket handlers. I'm on it now. Thanks for reviewing!

src/main/java/spark/SparkBase.java
perwendel: Ok, do what you find is best regarding the handlers.

nacx:
@perwendel Done. I've added a new commit to facilitate the review. It it looks good I'll squash the commit for a cleaner merge if needed.

perwendel:
:+1: 
Good stuff man! Thanks a lot!

nacx:
Thanks! Should #196 and #65 be marked as fixed now?

Also, I've seen the formatting follow-up commit. Is there any style guide or config I can import for future contributions?

vemacs:
@nacx I'm unable to get even the provided examples working on a standalone server. I can confirm that they're being registered (see https://github.com/perwendel/spark/issues/345#issuecomment-137273915), but any attempts to connect to them using a standard browser fails, and I get `The requested route [/echo] has not been mapped in Spark` on the server console.

PR #327, opened 2015-08-14
Add null check when Accept-Encoding header is empty.

briannesbitt:
Long story short...  This patches a possible NPE when the Accept-Encoding header is empty.

On a simple Hello World server with ```get("/", (request, response) -> "Hello World");``` try the following:

curl http://localhost:4567  <=== works fine
curl -H "Accept-Encoding;" -v http://localhost:4567   <=== NPE

There were other ways to fix the error (check if there is even an accept encoding header available etc. ) but this was the easiest.

### DIFF
--- a/src/main/java/spark/utils/GzipUtils.java
+++ b/src/main/java/spark/utils/GzipUtils.java
@@ -76,6 +76,10 @@ public class GzipUtils {
     private static class StringMatch implements Predicate<String> {
         @Override
         public boolean test(String s) {
+            if (s == null) {
+                return false;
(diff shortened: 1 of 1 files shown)


jkwatson:
Needs a unit test, IMO.

briannesbitt:
Funny thing is I tried to recreate it on the current master but couldn't.  Tried it on the current 2.2 (which is where I found the issue) and was able to re-create it.

Seems as though the update to jetty 9.2.11 helped fix this issue on the current master.

PR #333, opened 2015-08-19
Do not reset filterMatch in after filters fixed #332

### LINKED ISSUE #332: Path params are not available in the after filters
Let's say I have the following case, and I call the localhost:8080/hi/john.

While params are available in the route, they are not in the after filter. Is this normal?

``` java
    get("/hi/:name", (request, response) -> {
        System.out.println(request.params("name")); //Prints john
        return "hello";
    });


    after((request, response) -> {
        System.out.println(request.params("name")); //Prints null
    });
```

kliakos:
Request must carry all initial data in after filters

### DIFF
--- a/src/main/java/spark/http/matching/AfterFilters.java
+++ b/src/main/java/spark/http/matching/AfterFilters.java
@@ -45,8 +45,6 @@ final class AfterFilters {
                 if (context.requestWrapper().getDelegate() == null) {
                     Request request = RequestResponseFactory.create(filterMatch, context.httpRequest());
                     context.requestWrapper().setDelegate(request);
-                } else {
-                    context.requestWrapper().changeMatch(filterMatch);
                 }
 
                 context.responseWrapper().setDelegate(context.response());


Ariel-Isaacm:
I checked the problem and reproduced it, it seems your PR works fine, and all the tests pass but can this affect the application in some other way?

kliakos:
@Ariel-Isaacm as you said this is a "problem". A bug IMO.

I can't think of any way that it will affect the application, but even if it does, the framework will work as expected then.

Ariel-Isaacm:
I looked into the commits history and the line was added by @perwendel  with id 918842f, the reason was "Fix for #201, making body available" maybe @perwendel has a suggestion the best solution for this issue

perwendel:
I'll have a look this week. I've been busy lately so there are quite a few PR's to handle.

perwendel:
I have a hard time understanding the reason for this PR and I'm pretty sure i breaks functionality. The `changeMatch` updates the `params` and `splat` variables in the request which is fully related to how the request was matched for a specific route or filter. An example to explain:

```java
get("/foo/:a", ...
after("/foo/:b", ...
```

A request made on `/foo/stuff` in the handler provided to the `get` will have one entry in `params` with key `":a"` and in the handler provided to `after` will have one entry in `params` with key `":b"`. If `changeMatch` is not called this won't work.

I will close this and if I'm wrong, please explain and re-open.

kliakos:
In your example, what if the after filter is wildcard like this:

`after("/",)`

Then you have lost the information for the path params of the original request.

PR #334, opened 2015-08-24
no more sleeps in tests

amarseillan:
No more

```java
Thread.sleep(500)
```

in tests, making them deterministic (once and for all)


perwendel:
:+1:

PR #338, opened 2015-08-27
Add @FunctionalInterface to Spark's functional interfaces in its API

bentolor:
The java annotation `@FunctionalInterface` is an informal annotation type,  to indicate that an interface type declaration is intended to be a functional interface as defined by the Java Language Specification.

With this annotation the compiler will throw an error, if this interface is changed in a way which would breaking existing Lambda usages etc.

Upon studying the Sparks API I was stumbling over compiler errors and was missing these annotations as a reassuring documentation.


perwendel:
:+1:

PR #341, opened 2015-08-29
Add the ability to specify http only in a cookie

evertvp:
This is particularly useful for [preventing some XSS attacks](https://www.owasp.org/index.php/HttpOnly).


perwendel:
Should there not be a method where you don't have to set path to make it consistent?

evertvp:
I have added a method where you don't have to set the path.

perwendel:
:+1:

PR #342, opened 2015-08-29
Client-side session in cookie implementation

Lightraven:
Hey, I abstracted away the session in the framework and added a client-side session state. This can be very useful in distributed environments. 
With this implementation it should be very easy to add other session implementations, depending on the users requirements.

Please let me know what you think of it, if this is the way you would like to see it in the framework. One thing I am not so sure about is if the place where I write the session to the cookie is chosen correctly, because I have to set it after all processing as well as when a redirect occurs.

There are also some smaller improvements performance-wise.


cwildfoerster:
I would love to see that PR merged because i was thinking about implementing this myself, but i think the PR contains to many other changes or refactors? And personally, i dont like "IName" interfaces :)

Lightraven:
The added dependency is for fast serialization of the session data as it turned out that the JDK-implementation had some performance issues in my tests. Maybe we should drop the actual client-side session implementation to the tests folder, so we do not need the dependency?
Yes, I did some other small changes/optimizations, should I revert them? Other refactorings were neccessary to make other cookie-implementations work without the framework accessing the server session

perwendel:
Thanks for contributing!

I had a brief look and I'd prefer that PRs are specific since it makes it very hard to review if too many different changes have been made.

My compentence regarding cookies are limited. If you had to guess, how many Spark users would gain from this functionality addition? E.g. would it be like ~0.1% or 10%?

If this functionality is to be added then the public API method names etc. need to follow Spark conventions. Ie. not `set` and `get` prefixes for methods. No `I` prefix for interfaces.

PR #344, opened 2015-09-01
Re-write and refactoring of implementation.

perwendel:
The code has needed refactoring for a while. Functionality in "fat" classes has been moved to own components etc.

This PR also includes the removal of the global states in the Spark class in favor of an instance based implementation. This is needed for the planned future functionality for being able to have multiple Spark "servers" running in a single JVM.

There has also been changes to remove cyclic dependencies that has been introduced during the projects lifetime. There is one left that is extremely difficult to get rid of.

Inspired by #308. Thanks for the contribution.


src/main/java/spark/Routable.java
```diff
@@ -0,0 +1,759 @@
...
+    public void patch(String path,
+                       String acceptType,
+                       Route route,
+                       ResponseTransformer transformer) {
+        addRoute(HttpMethod.patch.name(), ResponseTransformerRouteImpl.create(path, acceptType, route, transformer));
+    }
+
+    // TO BE IMPLEMENTED
+
+    protected abstract void addRoute(String httpMethod, RouteImpl route);
```
perwendel: missing javadoc.

src/main/java/spark/Routable.java
```diff
@@ -0,0 +1,759 @@
...
+
+    protected abstract void addFilter(String httpMethod, FilterImpl filter);
```
perwendel: missing javadoc.

src/main/java/spark/Routable.java
```diff
@@ -0,0 +1,759 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package spark;
+
+import spark.route.HttpMethod;
+import spark.utils.SparkUtils;
+
+/**
+ * Routable interface. Lets implementing classes inherit default routable functionality.
```
perwendel: incorrect javadoc.

src/main/java/spark/Spark.java
```diff
@@ -857,4 +857,181 @@ public static ModelAndView modelAndView(Object model, String viewName) {
         return new ModelAndView(model, viewName);
     }
 
-}
+    ///////////////
+    // SparkBase //
```
perwendel: remove comment.

src/main/java/spark/Spark.java
```diff
@@ -857,4 +857,181 @@ public static ModelAndView modelAndView(Object model, String viewName) {
...
+    }
+
+    /**
+     * Stops the Spark server and clears all routes
+     */
+    public static synchronized void stop() {
+        getInstance().stop();
+    }
+
+    // WEBSOCKETS TODO:
```
perwendel: remove comment.

src/main/java/spark/Spark.java
```diff
@@ -36,11 +31,24 @@
  *
  * @author Per Wendel
  */
-public final class Spark extends SparkBase {
+public final class Spark {
```
perwendel: remove synchronization since the methods in SparkInstance are synchronized.

src/main/java/spark/SparkInstance.java
```diff
@@ -320,116 +320,115 @@ public static synchronized void stop() {
...
-     * Wraps the route in RouteImpl
-     *
-     * @param path  the path
-     * @param route the route
-     * @return the wrapped route
-     */
-    protected static RouteImpl wrap(final String path, final Route route) {
-        return wrap(path, DEFAULT_ACCEPT_TYPE, route);
-    }
+    // TODO: BEGIN
```
perwendel: remove obsolete comments.

src/main/java/spark/SparkInstance.java
```diff
@@ -276,11 +276,11 @@ public static synchronized void webSocket(String path, Class<?> handler) {
      *
      * @param timeoutMillis The max idle timeout in milliseconds.
      */
-    public static void webSocketIdleTimeoutMillis(int timeoutMillis) {
+    public void webSocketIdleTimeoutMillis(int timeoutMillis) {
```
perwendel: sync? more methods?

src/main/java/spark/globalstate/ServletFlag.java
```diff
@@ -0,0 +1,40 @@
...
+ * limitations under the License.
+ */
+package spark.globalstate;
+
+/**
+ * Holds the global information if Spark was run from an "external" web application server.
+ */
+public class ServletFlag {
+
+    private static boolean isRunFromServlet = false;
```
jkwatson: If this is truly "global state", should you use an AtomicBoolean here, to protect against possible thread access races?

src/main/java/spark/staticfiles/ServletStaticFiles.java
```diff
@@ -0,0 +1,109 @@
...
+    public static List<AbstractResourceHandler> staticResourceHandlers() {
+        return staticResourceHandlers;
+    }
+
+    /**
+     * Configures location for static resources
+     *
+     * @param folder the location
+     */
+    public static void configureStaticResources(String folder) {
```
tipsy: Flip some ifs and return early?

src/main/java/spark/globalstate/ServletFlag.java
perwendel: I'll change it. If you have time it would be great if you reviewed the whole PR. A lot of changes....

src/main/java/spark/Spark.java
perwendel: done!

src/main/java/spark/globalstate/ServletFlag.java
jkwatson: I'll try to get to it tonight. There's a lot of changes, as you say.

src/main/java/spark/staticfiles/ServletStaticFiles.java
perwendel: done!

perwendel: more done!

src/main/java/spark/Routable.java
```diff
@@ -0,0 +1,772 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package spark;
+
+import spark.route.HttpMethod;
+import spark.utils.SparkUtils;
+
+/**
+ * Routable abstract class. Lets extending classes inherit default routable functionality.
```
jkwatson: If we're all-in on java 8, we could use an interface with default methods, rather than an abstract class.

src/main/java/spark/RouteWrapper.java
```diff
@@ -0,0 +1,95 @@
...
+import spark.FilterImpl;
+import spark.Request;
+import spark.Response;
+import spark.Route;
+import spark.RouteImpl;
+
+/**
+ * Utility methods for wrapping routes.
+ */
+class RouteWrapper {
```
jkwatson: I wish this had a more evocative name.  I know it's an internal detail, but I still don't quite understand what the purpose of "wrapping" a route is, even though I've perused the source code frequently.

src/main/java/spark/Spark.java
```diff
@@ -36,11 +31,24 @@
...
+    private static SparkInstance instance;
+
     // Hide constructor
     private Spark() {
     }
 
+    /**
+     * Initializes singleton.
+     */
+    private static class SingletonHolder {
```
jkwatson: What purpose does this class serve? Why not just keep the instance at the top level?

src/main/java/spark/globalstate/ServletFlag.java
```diff
@@ -0,0 +1,42 @@
...
+     * Tells the system that Spark was run from an "external" web application server.
+     */
+    public static void runFromServlet() {
+        isRunFromServlet.compareAndSet(false, true);
+    }
+
+    /**
+     * @return true if Spark was run from an "external" web application server.
+     */
+    public static boolean isIsRunFromServlet() {
```
jkwatson: This method name is terrible.  How about `isRunningFromServlet()`?

src/main/java/spark/globalstate/ServletFlag.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public class ServletFlag {
+
+    private static AtomicBoolean isRunFromServlet = new AtomicBoolean(false);
+
+    /**
+     * Tells the system that Spark was run from an "external" web application server.
+     */
+    public static void runFromServlet() {
+        isRunFromServlet.compareAndSet(false, true);
```
jkwatson: If you're not using the return value of `compareAndSet`, then just `set` will do just fine.

jkwatson:
I've read your objections to an instance-based API, and I disagree very strongly with them. So much so, that I'm considering forking and maintaining my own version of Spark that is instance-based. IMO, modern java programming needs to utilize immutability to simplify interactions with multithreaded applications on multicore systems. Having shared, mutable state in the form of static state embedded in multiple classes is a recipe for maintenance and debugging hell. Modern java programmers have no trouble creating an instance in order to do work. The singleton nature of instances should be a concern of the application that uses them, not of the classes themselves.

The amount of extra work you have to do to maintain the initialization state of the SparkInstance should be a giant red flag that immutability would be a simpler and safer programming model.

I think that you could approach the static API differently *and* gain the advantages of immutability with something like:

```java
Spark.bootstrap()
     .port(8100)
     .get("/path", (req, res) -> "OK")
     .put("/path", new PutRoute())
     .post("/path", new Poster())
     .ignite();
```

With that model, you still get a pure "static" API, but one that reads fluently, and can maintain immutability in the result of the `.ignite()` call.

Given that you seem to be determined to stick with your current static API model, this seems like an ok PR, and a decent set of refactorings, with the several comments that I provided.

src/main/java/spark/Routable.java
perwendel: I did start with that but I fail(ed) to see how that would make things better in this case.

src/main/java/spark/RouteWrapper.java
perwendel: I agree. The name sucks, I'll change it.

I don't remember every little change that has been made. I believe it was to make the transition to the Spark 2.0 API smoother. This code could probably be made redundant by doing changes in SparkInstance.

src/main/java/spark/Spark.java
perwendel: It's one of many Singleton idioms. I normally use this one, that's the reason.
https://en.wikipedia.org/wiki/Singleton_pattern#Initialization-on-demand_holder_idiom

src/main/java/spark/globalstate/ServletFlag.java
perwendel: :), yes. Missed that.

src/main/java/spark/globalstate/ServletFlag.java
perwendel: I've never used AtomicBoolean before. compareAndSet will just update if it's 'false'? Set will always update? There's no difference in how this class will work so I guess set() is more logical.

perwendel:
@jkwatson Thanks for reviewing.

I'm trying to be as clear as possible because I don't think you really have understood what I am arguing for/against. I agree with a lot of what you're saying BUT this it not what I argue against.

I don't believe java developers have trouble creating instances. This is neither a problem nor a disqualifier.

I don't believe that would be more work maintaining an extra API. This is also neither a problem nor a disqualifier.

Having shared, mutable state in the form of static state embedded in multiple classes is a recipe for maintenance and debugging hell. This might be true. However, I haven't had any issues with this and the few unhappy developers mostly miss the functionality to run multiple spark servers in one JVM. The big majority of developers notice that Spark works almost perfect and are happy with that.

This is also not a conceptual argument having static methods or not like your code proposal (which btw is a very nice syntax).

The ONE and ONLY thing I'm arguing is the cleanness and "declarativeness" of the API and that I value that more than all the potential drawbacks from having it. Of course I can't be 100% sure that this is the right strategy but this is the position I have.

I value the clear API Spark has today more and I think changing the API or having multiple APIs would cost Spark more than the gain.

jkwatson:
I understand your position, and I respectfully disagree with it. I value the immutability and code clarity and safety over the sanctity of the existing API. You feel the opposite.

Even with my misgivings about the internals of Spark, it's still the best web framework in java there is, and I would still choose it over the hot mess that is JAX-RS and Jersey.  :)

perwendel:
I'm grateful that people with opinions raise them. It forces me to re-evaluate my own positions which only can be good for the framework and myself in the long run.

I'm also glad that you still find it to be the best web framework in Java.

Have a good one!

PR #349, opened 2015-09-06
fix route missing when an exception thrown

Tradunsky:
Fix following use case:

```java
exception(IllegalArgumentException.class,(exception, request, response) -> {
                    response.status(Response.SC_BAD_REQUEST);
                    response.body("It's bad idea");
                });

get("/throw", (request, response) -> {
            throw new IllegalArgumentException();
});
```

Expected: 
Response code: 400
Response body: "It's bad idea"

Actual:
Response code: 404
Response body: "<html><body><h2>404 Not found</h2></body></html>"

### DIFF
--- a/src/main/java/spark/webserver/MatcherFilter.java
+++ b/src/main/java/spark/webserver/MatcherFilter.java
@@ -16,33 +16,20 @@
  */
 package spark.webserver;
 
+import spark.*;
+import spark.route.HttpMethod;
+import spark.route.SimpleRouteMatcher;
+import spark.routematch.RouteMatch;
+import spark.utils.GzipUtils;
+import spark.webserver.serialization.SerializerChain;
+
+import javax.servlet.Filter;
+import javax.servlet.*;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.List;
-
-import javax.servlet.Filter;
-import javax.servlet.FilterChain;
-import javax.servlet.FilterConfig;
-import javax.servlet.ServletException;
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-
-import spark.Access;
-import spark.FilterImpl;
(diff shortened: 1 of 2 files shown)


perwendel:
I copy-pasted the code and it works for me without the fix. As you can see in the MatcherFilter code the `bodyContent` should have been set to `bodyAfterFilter` which later on would set `consumed` to true making `consumedAsException` redundant.

Am I missing something?

Tradunsky:
hm... It's really strange, because  the tests didn't success without this fix..

Sorry, I made mistake in use case:
```java
exception(IllegalArgumentException.class,
                (exception, request, response) -> {
                    response.status(Response.SC_BAD_REQUEST);
                });
        get("/throw", (request, response) -> {
            throw new IllegalArgumentException();
        });
```
After I've got message: 
"The requested route [/throw] has not been mapped in Spark",
and I debug code step-by-step, as u can see:
```java
String bodyAfterFilter = Access.getBody(responseWrapper.getDelegate()); //   bodyAfterFilter = null
if (bodyAfterFilter != null) { //false, then bodyContent wont assigned
                    bodyContent = bodyAfterFilter; 
}
boolean consumed = bodyContent != null; //false
if (!consumed && !isServletContext) { //true
            LOG.info("The requested route [" + uri + "] has not been mapped in Spark");
...
```

I think the deal with checking consumed by response body, but I really need to consume it without response body.

perwendel:
Ok, now I get it. I have to think about this. I guess even if body is not set the request should be seen as consumed if handled by an Exception handler!? The a fix would be (the else if):

```java

handler.handle(e, requestWrapper, responseWrapper);
String bodyAfterFilter = Access.getBody(responseWrapper.getDelegate());
if (bodyAfterFilter != null) {
      bodyContent = bodyAfterFilter;
} else if (bodyContent == null) {
      bodyContent = "";
}

```

right?

Tradunsky:
Actually, I don't think that u need to be focused on response body, because it can be empty or not and it is not depends on the processing of request.

Following to wiki:
https://en.wikipedia.org/wiki/List_of_HTTP_status_codes
there are some http codes which can be presented without body at all. For example, 204.

Tradunsky:
Maybe u need to declare some enum of request processing state or something like this.

PR #350, opened 2015-09-11
Disabling inputStream cache when Transfer-Encoding=chunked

sidneibjunior:
The current implementation for the HttpServletRequest returned from request.raw() (HttpRequestWrapper.class) caches the bytes when getInputStream() is called. When large data is transfered, an OutOfMemoryError is thrown, even when using chunked transfer encoding.


yuvalk:
Why is this not merged/closed?
Any other existing solution to receive large files, without cache blowing up memory?

TIA,
Yuval

tipsy:
@sidneibjunior could you update this?

tipsy:
Replaced by #821

PR #351, opened 2015-09-14
Route handling with enum of response state

Tradunsky:
According to pull request https://github.com/perwendel/spark/pull/349 for avoiding route missing when exception thrown and handled I've implemented enum of response state
It also fixed the error of processing response without content, e.g. 204

### DIFF
--- a/src/main/java/spark/Request.java
+++ b/src/main/java/spark/Request.java
@@ -16,23 +16,14 @@
  */
 package spark;
 
-import java.util.ArrayList;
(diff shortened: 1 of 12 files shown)


Tradunsky:
Not mapped in Spark jtest request detected ^.^

perwendel:
Please re-format according to https://github.com/perwendel/spark/tree/master/config for IntelliJ. This PR re-orders imports in incorrect order etc.

PR #358, opened 2015-09-28
copy stream directly rather than buffer as a utf-8 string

martylamb:
Hi Per,

This change copies a returned InputStream directly to output rather than read it in to memory first.  The InputStream might be large (making buffering a problem) and might not be valid UTF-8 (e.g. could be arbitrary data returned as application/octet-stream).


amarseillan:
:+1:

Tradunsky:
+1

perwendel:
:+1:

PR #359, opened 2015-09-28
Re-write implementation of static files.

### LINKED ISSUE #54: Custom 404 Page when using Static File Route?

perwendel:
Fixes #54 

Now the following works:

```java

staticFileLocation("/public");
        
get("/*", (q, a) -> {
    throw new NotFoundException();
});

exception(NotFoundException.class, (e, request, response) -> {
    response.status(404);
    response.body(NOT_FOUND_BRO);
});

```


src/main/java/spark/staticfiles/StaticFiles.java
```diff
@@ -16,56 +16,135 @@
...
+                LOG.error("Error when creating external StaticResourceHandler", e);
+            }
+            externalStaticResourcesSet = true;
         }
+
+    }
+
+    // PRIVATE //
+
+    private static boolean isStaticResourcesSet() {
```
jkwatson: if these are private, why not just refer to the actual fields?

src/main/java/spark/staticfiles/StaticFiles.java
```diff
@@ -16,56 +16,135 @@
...
+import spark.utils.IOUtils;
 
 /**
- * Utility methods for setting internal and external static file locations.
+ * Holds the static file configuration.
  */
 public class StaticFiles {
+    private static final Logger LOG = LoggerFactory.getLogger(StaticFiles.class);
+
+    private static List<AbstractResourceHandler> staticResourceHandlers = null;
```
jkwatson: All this static state makes me uncomfortable, as you've just made a blob of non-thread-safe global mutable state, that any client can manipulate.

perwendel:
Add support for ETAG and Cache-control headers?

PR #365, opened 2015-10-08
Do not force a logging implementation by default

nacx:
Spark is used in applications as a dependency and should not preconfigure the logging framework to be used. It should be up to the applications to pick their preferred SLF4J implementation without having to exclude the Spark's default impl when declaring the Spark dependency.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,12 +41,6 @@
             <version>1.7.12</version>
         </dependency>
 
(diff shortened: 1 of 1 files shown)


nacx:
Any thought on this?

perwendel:
duplicate of #381

perwendel:
fixed.

PR #371, opened 2015-10-20
Some refactors done

Ariel-Isaacm:
-AbstractResource.java  Added some Try-with-resource
-SimpleRouteMatcher.java {  
"Arrays.asList(routeEntry.acceptedType)" Always returns a list with a single element, so a SingletonList is the best way to go
"new RouteMatch(routeEntry.target, routeEntry.path, path, acceptType)" acceptType is always null
}
-GzipUtils.java simplified return statement
-IOUtils.java Redundant statement
-MimeParse.java Changed to a StringBuilder and optimized the append statements
- JettySparkServer.java webSocketServletContextHandler null check is done before, so there's no point checking it again.
  -StaticFilesTest.java since tmpExternalFile.delete(); returns a boolean if the File was deleted, I added some more information

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ public class SimpleExample {
 
 -------------------------------
 
-A simple CRUD example showing howto create, get, update and delete book resources
+A simple CRUD example showing how to create, get, update and delete book resources
 
 ```java
 import static spark.Spark.*;
@@ -108,7 +108,7 @@ import java.util.Map;
 import java.util.Random;
 
 /**
- * A simple CRUD example showing howto create, get, update and delete book resources.
+ * A simple CRUD example showing how to create, get, update and delete book resources.
  */
 public class Books {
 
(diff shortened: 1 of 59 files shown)


perwendel:
Tx for contrib. 

Before proper reviewing the following needs to be fixed:
* import order and policy don't comply with Spark standars. 
* Code formatting is not compliant with Spark standards in some places.

See: https://github.com/perwendel/spark/tree/master/config for intellij formatter.

Ariel-Isaacm:
Hi @perwendel 
When i try to execute the jar tells me that the jar file is corrupted any thoughts on this?

perwendel:
@Ariel-Isaacm 
You should import it under 'Import settings' in intellJ menu.

Ariel-Isaacm:
I applied the formatter to the code and it formatted more files that the ones i modified, i think it shouldn't have any trouble.

PR #381, opened 2015-11-02
Moved slf4j-simple to test scope

fzakaria:
Most users of the framework will BYOL (Bring Your Own Logger).
Having SLF4J-SIMPLE just causes a headache since they would need
to explicitly remove that as a transitive dependency.

Moving it to the test scope resolves this problem.


bentolor:
+1 for »BYOL« :-)

cwildfoerster:
+1

perwendel:
fixed.

PR #382, opened 2015-11-04
Make SparkInstance and Routable public.

hadashi:
Hi I just wanted to submit this PR is a means to talk about being able to control the initialisation of spark when using DI. I'm not strong on Java API design so maybe making these public is the wrong way to go about it. Anyway feedback is welcome.

If this is something that might be committable, I would add examples and possibly tests.

It would let us do something like below. 

```java
public abstract class ViewRoute eextends Routable {
    private SparkInstance instance
    public ViewRoute(SparkInstance instance) {
        this.instance = instance;
    }

    @Override
    public void addRoute(String httpMethod, RouteImpl route) {
        instance.addRouter(httpMethod, route);
    }

    @Override
    public void addFilter(String httpMethod, FilterImpl filter) {
       instance.addFilter(httpMethod, filter);
    }
}

@Singleton
public class LoginRoute extends ViewRoute {    
    @Inject
    public LoginRoute(SparkInstance instance) {
       super(instance);

       get("/login", (request, response) --> ....);
   }
}
```

### DIFF
--- a/src/main/java/spark/Routable.java
+++ b/src/main/java/spark/Routable.java
@@ -22,7 +22,7 @@ import spark.utils.SparkUtils;
 /**
  * Routable abstract class. Lets extending classes inherit default routable functionality.
  */
-abstract class Routable {
+public abstract class Routable {
 
     /**
      * Adds a route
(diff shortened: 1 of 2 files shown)


perwendel:
Hi and thanks for contributing.

This is a discussion that's been happening before, many times. Currently there is no plan to make an instance API for Spark. The reason behind that is solely based on API design.

If multiple endpoints (servers) functionality is added, then an instance API maybe the only way but we're not there yet.

Issue #367, opened 2015-10-15
Add new websocket method overload


soundvibe:
Currently, mapping path to websocket handler is done using this static method:

```
public static void webSocket(String path, Class<?> handler) {
    getInstance().webSocket(path, handler);
}
```

It works but the downside is that WebSocketCreatorFactory is now responsible for creating handler instance and no argument constructor should be used for the handler class. 
It would great if I could set already instantiated handler, so a simple manual DI through constructor would be possible.

The new method could look something like this:

```
public static void webSocket(String path, Object handler) {
    getInstance().webSocket(path, handler);
}
```

And example code using it could be something like this:
`webSocket("/status", new StatusWebSocket(dependency1, dependency2));`

jakaarl:
I'm reviewing different libraries and frameworks for an application utilizing WebSockets, and this is **exactly** what I found problematic with Spark, which I would otherwise love to use.

jakaarl:
I looked into this a bit, hoping it would be trivial to do. However, it's a bit complicated: handlers are first added to a path-to-handlerclass map and only instantiated when starting JettySparkServer. The simple solution would be to instantiate eagerly to avoid dealing with both classes and object instances. I wonder if there is a valid reason for the delayed instantiation?

tipsy:
Nice work!

PR #383, opened 2015-11-05
Enable passing WebSocket handler instances.

jakaarl:
Added a Spark.webSocket() method which accepts a handler instance
instead of a class. See issue #367.


jakaarl:
This is a quickly hacked up change I made and very briefly tested. Definitely needs to be reviewed by someone "in the know" and properly tested.

perwendel:
Looks good. Will review it more thoroughly ASAP.

jvtrigueros:
This would be really awesome to have! I'm currently running into this issue.

jakaarl:
I've looked a bit more into the code. The whole `WebSocketCreatorFactory` and `WebSocketCreator` thing does seem like unnecessary complexity, but since the latter is required by Jetty, there isn't that much to be achieved by refactoring. Also, my changes don't change thread safety of the web socket handlers: they are instantiated beforehand (at `ignite` before the changes and immediately after my changes) and the same instance serves all requests to the path.

jpthiery:
Hi,
I encount the same issue. I need to have interaction with other collaborator of my WebSocket instance which need to get arguments in it.

Thanks for your help.

cbeust:
It looks like this didn't get merged? (not seeing it in `spark-core:2.5`). Why not?

It would be really useful to specify the instance of the websocket handler instead of the class (in my case, this instance is injected by Guice so I want to pass my own instance instead of Spark calling `newInstance()` on the class.

jakaarl:
I drew the conclusion that contributions are not welcome after my PR was in limbo for four months or something to that effect. Closed the PR.

cbeust:
That's unfortunate :-(

PR #391, opened 2015-11-18
Code quality fix - Strings literals should be placed on the left side when checking for equality

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule squid:S1132 - “Strings literals should be placed on the left side when checking for equality”. 
You can find more information about the issue here: https://dev.eclipse.org/sonar/rules/show/squid:S1132

Please let me know if you have any questions.

Faisal.

### DIFF
--- a/src/main/java/spark/route/RouteEntry.java
+++ b/src/main/java/spark/route/RouteEntry.java
@@ -51,14 +51,14 @@ class RouteEntry {
                 String thisPathPart = thisPathList.get(i);
                 String pathPart = pathList.get(i);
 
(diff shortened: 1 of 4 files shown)


perwendel:
Thanks for contributing.

I would agree to the change if the change was from:
var != null && var.equals("literal")

In these case the variables/parameters should not be null. If null, there's a bug somewhere and it should not just be "swallowed" by the system.

Your thoughts!?

faisal-hameed:
Hi @perwendel. You are right that NPE should not be swallowed by system and it should be taken care explicitly. I tried to improve the only parts which does't effect the logic.

PR #352, opened 2015-09-14
Exception and route handle fixed: #349 

Tradunsky:
According to pull request #349 for avoiding route missing when exception thrown and handled I've implemented enum of response state.
It also fixed the error of processing response without content, e.g. 204

dev note:
broken jtests fixed too

PR #405, opened 2015-11-25
Fix route missing when exception thrown and handled/unhandled.

Tradunsky:
It also fix the error of processing response without content, e.g. 204

According to Per's request (https://github.com/perwendel/spark/pull/352#issuecomment-149323246) for PR #349 reformat code using spark's code style.

Also, reformat some log messages in MatcherFilter because it's really useful for the issue.


Tradunsky:
ping @perwendel

PR #408, opened 2015-11-26
Removed jetty utils URIUtil references from resource handling classes (fixes #407)

### LINKED ISSUE #407: jetty-util dependency in resource handling classes
AbstractResourceHandler.java has a addPaths function copied from jetty's URIUtil class but the getResource function still calls the URIUtil original function. When ran into an application server without the jetty dependencies, a ClassNotFoundException is raised. Same goes with ClasspathResourceHandler and the URIUtil.canonicalPath function, but this one hasn't been copied from URIUtil.

I think that jetty dependencies should be removed from this part of the framework.

MouettE-SC:
Fixes #407


perwendel:
Tx

Issue #221, opened 2014-12-11
Can not access response from routes in the after filters


kliakos:
It would be nice (not to say necessary) if response returned from route could be accessed with response.body() in the after filters.

This would enable easy response logging for example.

Otherwise, what is the purpose of the response.body() getter method?

I have already made this PR #211 that takes care of this issue.

aldendaniels:
Thanks @kliakos! I would love to see the PR merged.

Per the docs:
> After filters are evaluated after each request and can read the request and read/modify the response

Without this fix, we currently can't "read" the response body.

kliakos:
@aldendaniels  I had missed that notice in the docs.

This means that reading the response is the desired behaviour, not a change I propose. The current behaviour is definitely a bug as it is right now.

To deal with this I use my own custom build over the official one, since I see no progress overall in the project.

angexis:
No flame war intended with my PR #412 ... I just thought that the PR #211 had been around for over a year now and tried to figure out if there was a reason for it stagnation... Any of the two PRs reaching the next release would be really very nice! For me, this a kind of blocker...

kliakos:
No problem @angexis, as long as it does the trick. I really hadn't thought the case of the response body being binary or stream. Your solution seems much more robust, however there are much more changes also for the admins to review.

kliakos:
Issue has been fixed in the latest commits.

angexis:
To be honest, I have closed #412 w/o testing because my project no longer uses Spark... Cheers

tipsy:
Fixed in 2.5

PR #211, opened 2014-11-13
Make bodyContent available in the after filters

kliakos:

PR #412, opened 2015-12-03
Another (more drastic) way to fix issue #221.

angexis:
"After filters" have only limited usage if you cannot read the payload.
There is already PR #211 trying to fix the same issue... in a different manner.

I am open for any respectful discussion on why one method would be better or worse than the other. 

My goals were to:
1. simplify the code, as well supporting the general cases:
2. the payload in not a string (and you still want to modify it in a filter)
3. you want to chain the filters (and see the changed values)

Since, I am new to Spark, I hope I don't break any logic here...

Notice that I changed the "API" to add the 'payload' to the Response (for the general case where the body would not be a simple string).

I also added a test case BTW...

Cheers

### DIFF
--- a/src/main/java/spark/Access.java
+++ b/src/main/java/spark/Access.java
@@ -22,11 +22,7 @@ public final class Access {
 
     private Access() {
     }
-
-    public static String getBody(Response response) {
-        return response.body();
-    }
-
+    
     public static void changeMatch(Request request, RouteMatch match) {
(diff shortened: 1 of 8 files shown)


perwendel:
thanks for contributing. I'll have a look at this and #221 and try to figure out the best solution.

angexis:
No need to pull if the issue is already addressed (which I did not check)

PR #424, opened 2015-12-18
Support for removing registered routes

alonsod86:
These changes provide the functionality to remove already registered routes from the Spark singleton. Let's say that I have

Spark.get("/api", (request, response)->{...});

I could do

Spark.removeRoute("get", "api");

### DIFF
--- a/src/main/java/spark/Spark.java
+++ b/src/main/java/spark/Spark.java
@@ -986,6 +986,24 @@ public final class Spark {
         getInstance().stop();
     }
 
+    /**
+     * Removes the routes associated to the provided httpMethod and path
+     * @param httpMethod
+     * @param path
+     */
+    public static void removeRoute(String httpMethod, String path) {
+    	getInstance().removeRoute(httpMethod, path);
+    }
+    
+    /**
+     * Removes all the routes associated to the provided path
+     * @param httpMethod
+     * @param path
+     */
+    public static void removeRoute(String path) {
+    	getInstance().removeRoute(path);
+    }
+    
     ////////////////
     // Websockets //
 
--- a/src/main/java/spark/SparkInstance.java
+++ b/src/main/java/spark/SparkInstance.java
@@ -323,6 +323,26 @@ final class SparkInstance extends Routable {
         routeMatcher.parseValidateAddRoute(httpMethod + " '" + route.getPath() + "'", route.getAcceptType(), route);
     }
 
+    /**
+     * Removes the routes associated to the provided httpMethod and path
+     * @param httpMethod
+     * @param path
+     */
+    public void removeRoute(String httpMethod, String path) {
+    	init();
+    	routeMatcher.removeRoute(path, httpMethod);
+    }
+    
+    /**
+     * Removes all the routes associated to the provided path
+     * @param httpMethod
+     * @param path
+     */
+    public void removeRoute(String path) {
+    	init();
+    	routeMatcher.removeRoute(path);
+    }
+    
     @Override
     public void addFilter(String httpMethod, FilterImpl filter) {
         init();


3tar:
Just curious what would be the use case for such feature?

ruurd:
Oh goodie! An API that has indeterminate state.

alonsod86:
I have an use case that requires starting and stopping REST endpoints on different threads on the same JVM. This threads have complex functionality, but once they have done their task, I'd like to destroy the REST endpoint thath I previously started with it.

ruurd:
@alonsod86 you're not going to tell me that those REST endpoints are stateful somehow...

alonsod86:
No, of course they are not. It is just a matter of dynamic REST endpoints, i.e I'd like to publish endpoints such as /0/awesome, /1/awesome, /2/awesome, etc

Each of those "numbers" are threads running in my app.

ruurd:
Why then remove the path from the router? Make the thread number dynamic foo/{tid}/awesome and return a 404 if not found.

alonsod86:
Because the handler associated to foo/{tid}/awesome is still referenced in memory and attending requests after the thread is destroyed (Spark is a singleton instance for the entire app). This is a security issue, I just can't forget about those handlers.

ruurd:
Check if the thread exists possibly making sure thread numbering is consecutive if you do not want to reuse indexes. I don't see the security risk. I think you can do this exactly the same as looking up a database record by ID. No record => 404. No thread => 404, literally "I can't find that thread"

ruurd:
Check if the thread exists before accessing it. Similar to looking up a record from a database by its ID. No record => 404. In your case, no thread => 404. Not found. Logical.

alonsod86:
I am not sure if I understand your point of view. This would be a real use case: Say we have 2 threads consuming messages from the Twitter Stream API (each one with different credentials). While these threads are up and running they also expose an endpoint such as GET /{id}/search. Once the first thread is done with it's work, /0/search must be destroyed and unrefrenced, otherwise someone could keep making requests to that Twitter API endpoint (and it's corresponding credentials). 

From my point of view, this is not only a security risk but also a waste of resources, since we cannot reuse threads.

ruurd:
get("/:id/awesome", (request, response) -> {
            Thread t = ThreadManager.getThreadById(request.params(":id"));
            if (t != null) {
                t.doSomethingAwesomeWithTheThread();
            } else {
                response.status(404); // 404 Not found
                return "Thread not found";
            }
        });

ruurd:
Simple. Effective. No need to remove anything from the router. No dynamic creation or deletion of handlers needed.

alonsod86:
Well, that is "Simple and effective" because that handler is quite simple indeed! I am afraid our functionality is much more complex than just calling a method. The code that should run inside the REST handler is different on every thread, have different credentials depending on who invokes it, etc etc.

That is a good aproach for such a simple situation, but trust me on this, it does not fit for our requirements. It is ok if you don't like it, I am actually using a fork of this library on a clustered environment with hundreds of different threads and it's performance is.. for lack of a better word.. awesome!

ruurd:
Hah. Then you just need to add an extra level of indirection. Give every different piece of code its own Spark. Then pass through the invocation of the REST API from a Spark instance that only does that - forward the request to the proper Spark instance that can handle it.

And if you need to you can always add entry points to the front end spark to register and deregister any backend spark instances so the frontend knows in advance if a certain type is available.

tipsy:
It's tempting to merge this because it's such a small change, but I have to agree with @ruurd. It doesn't make sense (for 99.9% of users) to have a REST API with indeterminate state, so it probably shouldn't be a feature.

PR #429, opened 2015-12-29
instance should be reset while instance is stop.

qwIvan:
static variable should be initialize in next launch.
It can resolve some problem in testing.

### DIFF
--- a/src/main/java/spark/SparkInstance.java
+++ b/src/main/java/spark/SparkInstance.java
@@ -313,6 +313,11 @@ final class SparkInstance extends Routable {
             server.stop();
             latch = new CountDownLatch(1);
         }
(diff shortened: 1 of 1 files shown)


src/main/java/spark/SparkInstance.java
```diff
@@ -313,6 +313,11 @@ public synchronized void stop() {
             server.stop();
             latch = new CountDownLatch(1);
         }
+        sslStores = null;
```
dinesh707: extract a method call resetInstance(), and then do setting all those nulls there. It will improve readability

tipsy:
Closing this as it's very outdated. Feel feel to create a new PR if you still need this functionality.

PR #430, opened 2015-12-29
handle return null empty will rewrite to empty string

qwIvan:


tipsy:
This could have weird consequences for people switching Spark version, so we won't include in Spark 2.X.

PR #456, opened 2016-02-07
Add convenience methods for redirect

tipsy:


src/main/java/spark/Spark.java
```diff
@@ -780,6 +783,69 @@ public static void patch(String path,
...
+
+    /**
+     * Redirect from a verb-path to another path
+     *
+     * @param method   the http verb to redirect from
+     * @param fromPath the path to redirect from
+     * @param toPath   the path to redirect to
+     * @param status   the redirect status (3XX)
+     */
+    public static void redirect(HttpMethod method, String fromPath, String toPath, RedirectStatus status) {
```
perwendel: We should do something about `HttpMethod`. We probably don't want to make that part of the public API.

PR #464, opened 2016-02-15
Refactored EmbeddedServer creation Fixes #461

### LINKED ISSUE #461: jetty-server-*.jar required even when running on Tomcat
I used to exclude the jetty-server-*.jar for my Tomcat war build.

Latest version of Spark however has a problem with this, and throws ClassNotFoundException.

`Caused by: java.lang.NoClassDefFoundError: org/eclipse/jetty/server/Handler
        at spark.embeddedserver.EmbeddedServers.<clinit>(EmbeddedServers.java:37
)
        at spark.SparkInstance.<init>(SparkInstance.java:71)
        at spark.Spark$SingletonHolder.<clinit>(Spark.java:44)
        at spark.Spark.getInstance(Spark.java:48)
        at spark.Spark.get(Spark.java:564)
        ... 58 more
Caused by: java.lang.ClassNotFoundException: org.eclipse.jetty.server.Handler
        at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClas
sLoaderBase.java:1305)
        at org.apache.catalina.loader.WebappClassLoaderBase.loadClass(WebappClas
sLoaderBase.java:1157)
        ... 63 more`

kliakos:


perwendel:
Hi and thanks for contributing!

I'm leaning towards not merging this PR.

The small gain from this change is that `EmbeddedJettyFactory` is created on demand. It's a class with no members and really doesn't do anything until `create()` is invoked i.e. an instance with very small footprint. The loss is that the code is no longer straight-forward. Now, there's a mix of switch and specific Map functionality to solve something that is very trivial. For me it's not worth it for saving the creation of one instance.

Your thoughts?

kliakos:
The reason for this change is described in #461. It is not the performance gain, it is the fact that I do not want to have to incude any jetty related jar since I am deploying my war on Tomcat.

IMO this change does not make the code any more complex. I suppose it is about the way each one has code in his mind.

perwendel:
I missed the #461 part of this. My bad. Let me have another look.

I would definitely say the code is more complex. Having just a map is simpler than having a map and a switch combined. However, if the gain is big enough than that can be worth it. I'll check if it can be done in another way.

kliakos:
I refactored tha code to remove the switch statement, since this is your concern.

The only thing that is changed now is that instead of all embedded servers being created in the static constructor, they are added once in the create method.

However, if you have 5 embedded servers, with your approach you add them anyway without actually knowing if anyone will use them. Most likely someone will use only one.

perwendel:
I've created an alternative solution. Haven't had time to try it out since I don't have the infrastructure setup on the computer I'm on now. Please have a look #466

kliakos:
We did almost the same thing, moved the initialization to a method, instead of tha static constuctor.

Either way is fine.

perwendel:
Yup!

PR #466, opened 2016-02-16
Fix for #461, alternative to #464

perwendel:


kliakos:
Just tested this solution, it works fine.

PR #467, opened 2016-02-16
Option to not GZIP compress content

abradle:
New functionality to allow the user to specify that data should NOT be gzip compressed by spark itself.

If content-encoding and accept-encoding are gzip -> spark by default gzip compresses the response. This is a nice feature - however may not be desired (i.e. if content is already gzip compressed). I have added an option to not gzip compress the response (using a static boolean).

### DIFF
--- a/src/main/java/spark/utils/GzipUtils.java
+++ b/src/main/java/spark/utils/GzipUtils.java
@@ -37,7 +37,7 @@ public class GzipUtils {
     private static final String CONTENT_ENCODING = "Content-Encoding";
 
     private static final String GZIP = "gzip";
-
+    
     private static final StringMatch STRING_MATCH = new StringMatch();
 
     // Hide constructor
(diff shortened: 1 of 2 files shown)


src/main/java/spark/utils/GzipUtils.java
```diff
@@ -68,8 +68,10 @@ public static OutputStream checkAndWrap(HttpServletRequest httpRequest,
 
         if (acceptsGzip) {
             if (!requireWantsHeader || wantGzip) {
+            	if(spark.Spark.doGzip==true){
```
dinesh707: if you are accessing something in a some other class/object use a getter. IMO this is very confusing.

abradle: Thanks for the feedback! (I'm new to java programming). I agree with getters and setters general - but I was trying to mimic the way Spark does things more generally. So in the server setup you do not generate an instance of the class - so I thought that a static variable would be better here. (But I'm very happy to be told I'm wrong).

dinesh707: Im new to spark! no worries

tipsy:
@abradle this would need some sort of API-method like `disableAutomaticGzipping()`. Do you want to update the PR?

PR #468, opened 2016-02-17
Set content type for static file response

DNx5:
I keep getting this message from my browser's console
`Resource interpreted as Stylesheet but transferred with MIME type text/plain`
So, I made this fix.

### DIFF
--- a/src/main/java/spark/staticfiles/StaticFiles.java
+++ b/src/main/java/spark/staticfiles/StaticFiles.java
@@ -18,6 +18,7 @@ package spark.staticfiles;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


src/main/java/spark/staticfiles/StaticFiles.java
```diff
@@ -58,6 +59,10 @@ public static boolean consume(HttpServletRequest httpRequest,
             for (AbstractResourceHandler staticResourceHandler : staticResourceHandlers) {
                 AbstractFileResolvingResource resource = staticResourceHandler.getResource(httpRequest);
                 if (resource != null && resource.isReadable()) {
+                    String contentType = Files.probeContentType(resource.getFile().toPath());
```
dinesh707: shoudent there be a default content type ? and how come its null?

Issue #400, opened 2015-11-24
Tests fail if port 4567 is in use


tipsy:
Not sure if this is a bug, but we should probably use a random open port for tests. Or at least not 4567.

jakaarl:
From my quick test, seemed like at least these tests are affected:
 * `BodyAvailabilityTest`
 * `BooksIntegrationTest`
 * `CookiesIntegrationTest`
 * `GenericIntegrationTest`
 * `GenericSecureIntegrationTest`
 * `GZipTest`
 * `StaticFilesTest`

These are affected, because they use the default port. Others are affected when their corresponding ports are taken: for example, `ServletTest`binds to port 9393. Also, `SparkTestUtil` is relevant here, since it needs to be passed the Spark port - typical test pattern is to pass the default port in setup phase and not specify a port to Spark, causing it to default to 4567. I guess this pattern should be replaced by binding to an automatically assigned port (by passing in 0 as port argument) and then digging out the assigned port and pass it to `SparkTestUtil`. Problem is, that the port cannot (as far as I could tell with a quick glance at the sources) be easily pried out from `JettySparkServer`.

PR #470, opened 2016-02-18
 Tests fail if port 4567 is in use #400 

IRus:
I set port 4566 for tests. I think maybe we should use random port, but in this case we need access to server (connector) inside SparkTestUtil.

### DIFF
--- a/src/test/java/spark/util/SparkTestUtil.java
+++ b/src/test/java/spark/util/SparkTestUtil.java
@@ -42,7 +42,7 @@ import static spark.Spark.port;
 
 public class SparkTestUtil {
 
-^    public static final int DEFAULT_TESTS_PORT = 4566;
+    public static final int DEFAULT_TESTS_PORT = 4566;
 
     private int port;
 


src/test/java/spark/BodyAvailabilityTest.java
```diff
@@ -38,7 +38,7 @@ public static void tearDown() {
     public static void setup() {
         LOGGER.debug("setup()");
 
-        testUtil = new SparkTestUtil(4567);
+        testUtil = new SparkTestUtil(0);
```
dinesh707: can you add a constant explaining this 0 ?

IRus: Yes, i think i can add `public static int RANDOM_PORT = 0` inside `SparkTestUtil`

dinesh707: if thats really random, i would go with a inline comment, than a constant

IRus: It's not random. It's just mean that you doesn't care about port. I think even better to add default constructor for this case. Thank you for pointing!

dinesh707: thats a good point !..

IRus:
Since an year without any comments from @perwendel... close

PR #476, opened 2016-02-29
Close the static file input stream when done

paavohuhtala:
Currently, the method `spark.staticfiles.StaticFiles.consume` leaves the file input stream open. In addition to being a bad idea in general, it also means that the server process has a file lock on all the static files all the time, which makes changing them on the fly (say using `gulp watch`) impossible. This PR fixes the issue by extracting the call to `resource.getInputStream()` into its own local variable, and explicitly closing the stream.

### DIFF
--- a/src/main/java/spark/staticfiles/StaticFilesConfiguration.java
+++ b/src/main/java/spark/staticfiles/StaticFilesConfiguration.java
@@ -17,6 +17,7 @@
 package spark.staticfiles;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.Date;
@@ -92,11 +93,12 @@ public class StaticFilesConfiguration {
                         httpResponse.setHeader(MimeType.CONTENT_TYPE, MimeType.fromResource(resource));
                     }
                     customHeaders.forEach(httpResponse::setHeader); //add all user-defined headers to response
-                    OutputStream wrappedOutputStream = GzipUtils.checkAndWrap(httpRequest, httpResponse, false);
 
-                    IOUtils.copy(resource.getInputStream(), wrappedOutputStream);
-                    wrappedOutputStream.flush();
-                    wrappedOutputStream.close();
+                    try (InputStream inputStream = resource.getInputStream();
+                         OutputStream wrappedOutputStream = GzipUtils.checkAndWrap(httpRequest, httpResponse, false)) {
+                        IOUtils.copy(inputStream, wrappedOutputStream);
+                    }
+
                     return true;
                 }
             }


src/main/java/spark/staticfiles/StaticFiles.java
```diff
@@ -58,11 +59,11 @@ public static boolean consume(HttpServletRequest httpRequest,
...
                 AbstractFileResolvingResource resource = staticResourceHandler.getResource(httpRequest);
                 if (resource != null && resource.isReadable()) {
-                    OutputStream wrappedOutputStream = GzipUtils.checkAndWrap(httpRequest, httpResponse, false);
-                    IOUtils.copy(resource.getInputStream(), wrappedOutputStream);
-                    wrappedOutputStream.flush();
-                    wrappedOutputStream.close();
-                    return true;
+                    try (InputStream inputStream = resource.getInputStream();
+                         OutputStream wrappedOutputStream = GzipUtils.checkAndWrap(httpRequest, httpResponse, false)) {
+                        IOUtils.copy(inputStream, wrappedOutputStream);
```
nkavian: Now you have forgotten to flush/close the output stream.

nkavian: Oops, I missed the line wrapping.

tipsy:
@paavohuhtala Could you update this?

paavohuhtala:
@tipsy As far as I remember, the static files support was heavily refactored two months after this PR, and obviously no one merged this before before the refactor.

I haven't used Spark since then, but I just checked and the bug still exists, [here](https://github.com/perwendel/spark/blob/1d84ebc3f271cc01f44d0055d59ef093067bbb0d/src/main/java/spark/staticfiles/StaticFilesConfiguration.java#L97). I could probably update this patch, but if it takes an another year to get merged then I'm not going to bother.

tipsy:
@paavohuhtala I will merge it right now if you update it. I know a lot of PR's have been left for dead.. It sucks, and I'm going through them all now, trying to clean stuff up. I could push this directly myself, but I think it's more fair/proper to merge the PR created by a community member, even if it is a year late.

paavohuhtala:
Accidentally closed this while updating my fork. I'll have the update ready in a few mins.

paavohuhtala:
Okay, the updated code is in. I didn't write a unit/regression test for this, as it's a very simple change and testing it properly might have required a non-trivial amount of refactoring.

tipsy:
Great, thank you!

paavohuhtala:
I think it's pretty impressive this took exactly 365 days to get merged. 😄

tipsy:
We might not do things fast, but we do them in style 😎

PR #478, opened 2016-03-05
Adding support to set port externally with system property

cardil:
Unfortunately Spark doesn't support setting port from system property. It's sometimes nessesery to give port for example in Heroku https://devcenter.heroku.com/articles/java-faq#what-constraints-should-i-be-aware-of-when-developing-applications-on-heroku:

> The web process must listen on one and only one port. The port must be the one specified in the $PORT variable. If your process listens on other ports, it will be shut down by Heroku.

In Spring Boot you can give port from command line like this:

`java -Dserver.port=$PORT -jar spring-app-1.0.0.jar`

In WildFly Swarm you also can give port like this:

`java -Dswarm.http.port=$PORT -jar swarm-app-1.0.0.jar`

I'm proposing to solve this by using the same property name as in Spring Boot `server.port` for easier migration between those two.

This PR will also makes posible my feature for Gasper: https://github.com/wavesoftware/java-gasper/issues/1

I tested those changes with test application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ public class HelloWorld {
 
 View at: http://localhost:4567/hello
 
+You can change default Spark port using Java system property `-Dserver.port=54321` or by
+using method `port()`.
 
 Check out and try the examples in the source code.
 You can also check out the javadoc. After getting the source from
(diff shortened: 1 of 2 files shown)


jkwatson:
I've run spark apps on heroku a lot, and it's never been an issue to grab the PORT environment variable as a part of the Main method that you start your app with, and initialize Spark with it.

cardil:
@jkwatson You apps are using Heroku convention then. I'ts vendor locking. I'm proposing to use library solution, like Spring and Swarm have done.

tipsy:
@cardil Can you update this?

tipsy:
I'll close this then. Feel free to re-open if you want this added after all.

Issue #479, opened 2016-03-07
[Enchancement] Simplify route creation


bademux:
1. There are no way to create all routes then start instance - for each addRoute\addFilter  `init()` method is invoked.
   [SimpleRouteMatcher.parseValidateAddRoute](https://github.com/perwendel/spark/blob/master/src/main/java/spark/SparkInstance.java#L282)  get some string concatenation `httpMethod + " '" + route.getPath() + "'"` as argument and internally parse it with [validation](https://github.com/perwendel/spark/blob/master/src/main/java/spark/route/SimpleRouteMatcher.java#L64). IMHO, we should pass `spark.route.HttpMethod` and `url` as separate arguments 
2. Shouldn't spark.Routable be an interface with default method as [comment states](https://github.com/perwendel/spark/blob/master/src/main/java/spark/Routable.java#L44)
3. spark.route.HttpMethod enum naming convention is confusing: enum names (GET\POST...) should be uppercased.
4. spark.routematch.RouteMatch#target is type of java.lang.Object it destroys typesafety and complicates the code.

What do you think about this changes? If you ok with changes I can prepare commit.

bademux:
btw, what does mean ```acceptType``` in signature ``void get(String path, String acceptType, Route route)```? Is it response type?

bademux:
Exception can be simplified, too

spark.SparkInstance#exception
```java
    public synchronized void exception(Class<? extends Exception> exceptionClass, ExceptionHandler handler) {
        // wrap
        ExceptionHandlerImpl wrapper = new ExceptionHandlerImpl(exceptionClass) {
            @Override
            public void handle(Exception exception, Request request, Response response) {
                handler.handle(exception, request, response);
            }
        };

        ExceptionMapper.getInstance().map(exceptionClass, wrapper);
    }
```
In java8 it can be done in more elegant way, smth like
```java
ExceptionMapper.getInstance().map(exceptionClass, () -> handler.handle(exception, request, response));
```
so ExceptionHandlerImpl can be safely removed, although it will be breaking change

tipsy:
Sounds reasonable to me. @perwendel ?

mahmed8003:
```java
public class Routes implements ApplicationRoutes {

	@Override
	public void init(Router router) {

		router.BEFORE().with(BeforeLoggerFilter.class);

		router.GET("/").with(HomeRoute.class, JsonTransformer.class);
		router.GET("/test").with(EchoRoute.class, JsonTransformer.class);
[...]

PR #480, opened 2016-03-07
Type info for ExceptionHandler

JWGmeligMeyling:
Adding type information to the `ExceptionHandler`, so you do not have to cast the `Exception` to its subtype manually within the `handle` method.

### DIFF
--- a/src/main/java/spark/ExceptionHandlerImpl.java
+++ b/src/main/java/spark/ExceptionHandlerImpl.java
@@ -16,18 +16,19 @@
  */
 package spark;
 
-public abstract class ExceptionHandlerImpl implements ExceptionHandler {
+public abstract class ExceptionHandlerImpl<T extends Exception> implements ExceptionHandler<T> {
+
     /**
      * Holds the type of exception that this filter will handle
      */
-    protected Class<? extends Exception> exceptionClass;
+    protected Class<? extends T> exceptionClass;
 
     /**
      * Initializes the filter with the provided exception type
      *
      * @param exceptionClass Type of exception
      */
-    public ExceptionHandlerImpl(Class<? extends Exception> exceptionClass) {
+    public ExceptionHandlerImpl(Class<T> exceptionClass) {
         this.exceptionClass = exceptionClass;
     }
 
@@ -36,7 +37,7 @@ public abstract class ExceptionHandlerImpl implements ExceptionHandler {
      *
      * @return Type of exception
      */
-    public Class<? extends Exception> exceptionClass() {
+    public Class<? extends T> exceptionClass() {
         return this.exceptionClass;
     }
 
@@ -45,7 +46,7 @@ public abstract class ExceptionHandlerImpl implements ExceptionHandler {
      *
      * @param exceptionClass Type of exception
      */
-    public void exceptionClass(Class<? extends Exception> exceptionClass) {
+    public void exceptionClass(Class<? extends T> exceptionClass) {
         this.exceptionClass = exceptionClass;
     }
 
@@ -56,5 +57,5 @@ public abstract class ExceptionHandlerImpl implements ExceptionHandler {
      * @param request   The request object providing information about the HTTP request
      * @param response  The response object providing functionality for modifying the response
      */
-    public abstract void handle(Exception exception, Request request, Response response);
+    public abstract void handle(T exception, Request request, Response response);
 }
--- a/src/main/java/spark/Spark.java
+++ b/src/main/java/spark/Spark.java
@@ -879,7 +879,7 @@ public class Spark {
      * @param exceptionClass the exception class
      * @param handler        The handler
      */
-    public static void exception(Class<? extends Exception> exceptionClass, ExceptionHandler handler) {
+    public static <T extends Exception> void exception(Class<T> exceptionClass, ExceptionHandler<? super T> handler) {
         getInstance().exception(exceptionClass, handler);
     }
 
--- a/src/main/java/spark/ExceptionHandler.java
+++ b/src/main/java/spark/ExceptionHandler.java
@@ -4,7 +4,7 @@ package spark;
  * Created by Per Wendel on 2014-05-10.
  */
 @FunctionalInterface
-public interface ExceptionHandler {
+public interface ExceptionHandler<T extends Exception> {
 
     /**
      * Invoked when an exception that is mapped to this handler occurs during routing
@@ -13,5 +13,5 @@ public interface ExceptionHandler {
      * @param request   The request object providing information about the HTTP request
      * @param response  The response object providing functionality for modifying the response
      */
-    void handle(Exception exception, Request request, Response response);
+    void handle(T exception, Request request, Response response);
 }
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -540,11 +540,11 @@ public final class Service extends Routable {
      * @param exceptionClass the exception class
      * @param handler        The handler
      */
-    public synchronized void exception(Class<? extends Exception> exceptionClass, ExceptionHandler handler) {
+    public synchronized <T extends Exception> void exception(Class<T> exceptionClass, ExceptionHandler<? super T> handler) {
         // wrap
-        ExceptionHandlerImpl wrapper = new ExceptionHandlerImpl(exceptionClass) {
+        ExceptionHandlerImpl wrapper = new ExceptionHandlerImpl<T>(exceptionClass) {
             @Override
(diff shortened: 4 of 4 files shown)


src/main/java/spark/ExceptionHandlerImpl.java
```diff
@@ -16,18 +16,18 @@
  */
 package spark;
 
-public abstract class ExceptionHandlerImpl {
+public abstract class ExceptionHandlerImpl<T extends Exception> {
     /**
      * Holds the type of exception that this filter will handle
      */
-    protected Class<? extends Exception> exceptionClass;
+    protected Class<? extends T> exceptionClass;
```
gqio: I think since T extends already Exception, there is no need to have the wildcard here

JWGmeligMeyling: Thanks, changed the generics slightly!

src/main/java/spark/SparkInstance.java
```diff
@@ -327,11 +327,11 @@ public synchronized void init() {
      * @param exceptionClass the exception class
      * @param handler        The handler
      */
-    public synchronized void exception(Class<? extends Exception> exceptionClass, ExceptionHandler handler) {
+    public synchronized <T extends Exception> void exception(Class<T> exceptionClass, ExceptionHandler<? super T> handler) {
```
gqio: T is already the lower bound of the type you can have here, so I think "? super T" does not provide more information for the handler nor reduce the scope of the contravariance, don't you think so ? ExceptionHandle<T> should be fine

src/main/java/spark/ExceptionHandlerImpl.java
```diff
@@ -16,18 +16,18 @@
...
 
     /**
      * Initializes the filter with the provided exception type
      *
      * @param exceptionClass Type of exception
      */
-    public ExceptionHandlerImpl(Class<? extends Exception> exceptionClass) {
+    public ExceptionHandlerImpl(Class<? extends T> exceptionClass) {
```
gqio: same as above

src/main/java/spark/ExceptionHandlerImpl.java
```diff
@@ -36,7 +36,7 @@ public ExceptionHandlerImpl(Class<? extends Exception> exceptionClass) {
      *
      * @return Type of exception
      */
-    public Class<? extends Exception> exceptionClass() {
+    public Class<? extends T> exceptionClass() {
```
gqio: same as above

src/main/java/spark/ExceptionHandlerImpl.java
```diff
@@ -45,7 +45,7 @@ public ExceptionHandlerImpl(Class<? extends Exception> exceptionClass) {
      *
      * @param exceptionClass Type of exception
      */
-    public void exceptionClass(Class<? extends Exception> exceptionClass) {
+    public void exceptionClass(Class<? extends T> exceptionClass) {
```
gqio: same as above

src/main/java/spark/Spark.java
```diff
@@ -797,7 +797,7 @@ public static void patch(String path,
      * @param exceptionClass the exception class
      * @param handler        The handler
      */
-    public static void exception(Class<? extends Exception> exceptionClass, ExceptionHandler handler) {
+    public static <T extends Exception> void exception(Class<T> exceptionClass, ExceptionHandler<? super T> handler) {
```
gqio: same as below

src/main/java/spark/SparkInstance.java
JWGmeligMeyling: The `super` allows for a more general `ExceptionHandler` to be used for a more specific `Exception`. Example:

```java
static class MoreSpecificException extends RuntimeException {}

static class RuntimeExceptionHandler implements ExceptionHandler<RuntimeException> {
    @Override
    public void handle(RuntimeException exception, Request request, Response response) {
    }
}

@Test
public void testName() throws Exception {
     Spark.exception(MoreSpecificException.class, new RuntimeExceptionHandler());
}
```

src/main/java/spark/Spark.java
JWGmeligMeyling: Here it does not apply, as the `T` still has to extend `Exception`.

src/main/java/spark/ExceptionHandlerImpl.java
```diff
@@ -45,7 +45,7 @@ public ExceptionHandlerImpl(Class<? extends Exception> exceptionClass) {
      *
      * @param exceptionClass Type of exception
      */
-    public void exceptionClass(Class<? extends Exception> exceptionClass) {
+    public void exceptionClass(Class<T> exceptionClass) {
```
JWGmeligMeyling: Without the `? extends Exception` this setter has become effectively useless, as you can only set the exception type to the same type of exception. If we want to keep that behaviour, we should set the fields back to `? extends Exception` (as in my first commit). The generic type then specifies the most general type for this `ExceptionHandlerImpl` (and default to a default `Exception`), but the actual class may be more specific and set through the constructor or setter.

src/main/java/spark/ExceptionHandlerImpl.java
JWGmeligMeyling: Changed it back, because otherwise the exceptionClass get and set methods do not work as before using the default case (t = Exception)

bademux:
Proper generified implementation should look like this:
```java
public abstract class ExceptionHandlerImpl<T extends Class<? extends Exception>{

    protected Class<T> exceptionClass;

    public ExceptionHandlerImpl(Class<T> exceptionClass) {
        this.exceptionClass = exceptionClass;
    }
}
```

anyway, what is the purpose of this class?

JWGmeligMeyling:
No idea in what cases it would be useful actually. But as it is part of the public API it is probably best to leave it functioning the same as before.

bademux:
Ok, I see where it used:
```java
    public synchronized void exception(Class<? extends Exception> exceptionClass, ExceptionHandler handler) {
        // wrap
        ExceptionHandlerImpl wrapper = new ExceptionHandlerImpl(exceptionClass) {
            @Override
            public void handle(Exception exception, Request request, Response response) {
                handler.handle(exception, request, response);
            }
        };

        ExceptionMapper.getInstance().map(exceptionClass, wrapper);
    }
```
In java8 it can be done in more elegant way, smth like
```java
ExceptionMapper.getInstance().map(exceptionClass, () -> handler.handle(exception, request, response));
```

JWGmeligMeyling:
I would vote to remove it. Using the interface is more convenient than extending the `ExceptionHandlerImpl` anyway.

bademux:
@JWGmeligMeyling can you please point me to the place where you want to use Interface? I'm new to the spark, so may miss something.

JWGmeligMeyling:
I mean exactly the same way as you propose. So removing the `ExceptionHandlerImpl` and changing its usage to `ExceptionMapper.getInstance().map(exceptionClass, (exception, request, response) -> handler.handle(exception, request, response));`

bademux:
yep, it definitely worth to simplify it by using lambda style.

Will our changes fit into project vision of the Owner? Ping @perwendel 

btw, I'm collecting route simplification in  #479 You can already check some POC changes.

tipsy:
@JWGmeligMeyling Want to update this?

JWGmeligMeyling:
Yes I can rebase this

tipsy:
Pinging @JWGmeligMeyling

JWGmeligMeyling:
@tipsy Rebased! 😄

JWGmeligMeyling:
Not sure why travis fails, locally it passes.

```
Tests run: 290, Failures: 0, Errors: 0, Skipped: 0

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 46.415 s
[INFO] Finished at: 2017-04-09T16:07:09+02:00
[INFO] Final Memory: 21M/211M
[INFO] ------------------------------------------------------------------------
```

tipsy:
@JWGmeligMeyling Could you also add a test?

JWGmeligMeyling:
Aren't exceptions already tested? So the fact that it doesnt break with more explicit types should be enough ;)

tipsy:
Fair enough, I will add a small test in your honor though.

tipsy:
@JWGmeligMeyling and here it is (deb600532dc8536dcfa558bb5a23dddff60b2d68).
Thanks for the PR, sorry it took so long to merge!

JWGmeligMeyling:
Thanks for merging! 👍

PR #484, opened 2016-03-11
Optimized UriPath.java and created UriPath.test

thesnowgoose:
Moved @PrepareForTest to class header in WebSocketServletContextHandlerFactoryTest.java because running each test individually they failed. Each test should be independent of each other. 
Removed redundant casts.
Removed unnecessary variable in JettyServer.java

### DIFF
--- a/src/main/java/spark/resource/UriPath.java
+++ b/src/main/java/spark/resource/UriPath.java
@@ -1,18 +1,5 @@
 /*
  * Copyright 2016 - Per thesnowgoose
- *
- *  Licensed under the Apache License, Version 2.0 (the "License");
- *  you may not use this file except in compliance with the License.
- *  You may obtain a copy of the License at
- *
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


thesnowgoose:
@perwendel @tipsy is there any chance you can check this pull request?? I think I did it good, and the UriPath class is gonna run faster.

cesartg:
@thesnowgoose The UriPath code looks good! It is simpler and clearer that the copied code from Jetty. Also, you added the lacking unit test. Nice!

I see other files changes unrelated to the UriPath. I see they are minor improvements. They should be on other pull request but I am OK with them.

cc: @perwendel @tipsy

tipsy:
Looks good to me, but @perwendel has to have a look (he might be on easter vacation now).

src/main/java/spark/resource/UriPath.java
```diff
@@ -30,138 +20,43 @@
      * @return path or null.
      */
     public static String canonical(String path) {
-        if (path == null || path.length() == 0) {
+        if (path == null || path.isEmpty()){
```
dinesh707: here you can use StringUtils (apache commons) "isBlank" or "isEmpty".

thesnowgoose:
I did a minor change and improve the execution speed, also added another test case. @perwendel  It would be great if you could take a look.

tipsy:
I like this PR. Could you update to follow the coding-conventions? https://github.com/perwendel/spark/blob/master/config/spark-formatter-intellij.jar

tipsy:
Replaced by PR #792

PR #487, opened 2016-03-15
Re-added general exception logging in route/filter (fixes #486)

### LINKED ISSUE #486: No logging of uncaught exception in user code
Previous versions of spark would log uncaught exceptions in filter/routes especially when there is no exception handler but it doesn't log anything which makes it difficult to debug problems in user code during development. Based on what was done in 2.3 (adapted to the new codebase of 2.4) I will provide a small pull request to fix it.

MouettE-SC:


src/main/java/spark/http/matching/GeneralError.java
```diff
@@ -47,6 +49,7 @@ static void modify(HttpServletResponse httpResponse,
                 body.set(bodyAfterFilter);
             }
         } else {
+            LOG.error("", e);
```
dinesh707: i would still add some text for error message

juliaaano:
Wouldn't be better to log this in a finally block in _MatcherFilter.doFilter_? This way it would also log the caught _HaltException_. Besides, there's already a _Logger_ instance in that class, you wouldn't need to a new one in _GeneralError_.
I could submit a pull request if you let me.
Thanks.

gaojieliu:
When will this change be merged to master and release? We encounter the same issue and would like to have the fix soon.

MouettE-SC:
Updated PR to current tree

Issue #421, opened 2015-12-11
Routes having issues with /


SonarBeserk:
Spark Version: 2.3

``` java
        get("api/v1.0/test", (req, res) -> {
            return "String Test";
        });
```

When going to /api/v1.0/test it prints "String Test" as expected. However going to /api/v1.0/test/ causes a 404 error and a message in console.

```
spark.webserver.MatcherFilter - The requested route [/api/v1.0/test/] has not been mapped in Spark
```

franklinexpress:
Yup, happens to me too, been wondering about that. I think it sees it as a different route.

tipsy:
Duplicates #109

PR #488, opened 2016-03-16
removes last character from the uri string if it ends with slash

### LINKED ISSUE #474: Allow to match paths where the client send end or no ending slash
If you have a REST service that maps to path `user/list` then clients may call the service with context-path `user/list` or `user/list/` - eg notice the extra ending slash. 

It would be nice if you could configure spark-rest to allow to match both situations. Today only the former is matched.

You can see this in the logs as

```
[qtp1966741820-28] INFO spark.webserver.MatcherFilter - The requested route [/user/list/] has not been mapped in Spark
```

paulina-pacheco:
Fixes #474 and #421

### DIFF
--- a/src/main/java/spark/http/matching/MatcherFilter.java
+++ b/src/main/java/spark/http/matching/MatcherFilter.java
@@ -92,6 +92,10 @@ public class MatcherFilter implements Filter {
 
         String httpMethodStr = method.toLowerCase();
         String uri = httpRequest.getPathInfo();
+        if (uri.endsWith("/")){
+            uri=uri.substring(0,uri.length()-1);
+        }
+
         String acceptType = httpRequest.getHeader(ACCEPT_TYPE_REQUEST_MIME_HEADER);
 
         Body body = Body.create();


src/main/java/spark/http/matching/MatcherFilter.java
```diff
@@ -92,6 +92,10 @@ public void doFilter(ServletRequest servletRequest,
 
         String httpMethodStr = method.toLowerCase();
         String uri = httpRequest.getPathInfo();
+        if (uri.endsWith("/")){
```
dinesh707: can you add a test ?

And what happens if url = "hello/world?color=red&separator=/"

perwendel:
build failed.

perwendel:
The thing with this PR is that it both changes the behavior of Spark which would mean a version bump up to 3.0. The other thing is that I'm fairly sure that this would actually be incorrect behavior to have as default.

To address the issues #474 and #421 there should be a solution more like:
http://www.sinatrarb.com/faq.html#slash

PR #497, opened 2016-03-25
Change to class variable

t-kameyama:
I think the "p" (key regex pattern) variable should be a class variable.

### DIFF
--- a/src/main/java/spark/QueryParamsMap.java
+++ b/src/main/java/spark/QueryParamsMap.java
@@ -33,6 +33,8 @@ public class QueryParamsMap {
 
     private static final QueryParamsMap NULL = new NullQueryParamsMap();
 
+    private static final Pattern KEY_PATTERN = Pattern.compile("\\A[\\[\\]]*([^\\[\\]]+)\\]*");
+
     /**
      * Holds the nested keys
      */
@@ -42,8 +44,6 @@ public class QueryParamsMap {
      * Value(s) for this key
      */
     private String[] values;
-
-    private Pattern p = Pattern.compile("\\A[\\[\\]]*([^\\[\\]]+)\\]*");
 
     /**
      * Creates a new QueryParamsMap from and HttpServletRequest. <br>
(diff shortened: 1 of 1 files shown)


src/main/java/spark/QueryParamsMap.java
```diff
@@ -33,6 +33,8 @@
 
     private static final QueryParamsMap NULL = new NullQueryParamsMap();
 
+    private static final Pattern KEY_PATTERN = Pattern.compile("\\A[\\[\\]]*([^\\[\\]]+)\\]*");
```
dinesh707: can you add a doc comment about this pattern. Then its easy to read in future

src/main/java/spark/QueryParamsMap.java
```diff
@@ -33,6 +33,9 @@
 
     private static final QueryParamsMap NULL = new NullQueryParamsMap();
 
+    /**
+     * Pattern for parsing the key of querystring
+     */
     private static final Pattern KEY_PATTERN = Pattern.compile("\\A[\\[\\]]*([^\\[\\]]+)\\]*");
```
t-kameyama: @dinesh707 
I added javadoc comment

dinesh707:
+1

src/main/java/spark/QueryParamsMap.java
```diff
@@ -34,6 +34,11 @@
     private static final QueryParamsMap NULL = new NullQueryParamsMap();
 
     /**
+     * Pattern for parsing the key of querystring
+     */
+    private static final Pattern KEY_PATTERN = Pattern.compile("\\A[\\[\\]]*([^\\[\\]]+)\\]*");
```
perwendel: I'm not the author of this code but as my understanding goes the matcher that's fetched from the pattern instance has an internal state. This change means that all requests would share that state. Doesn't feel right.

t-kameyama:
@perwendel 
The Patten class is thread safe.
The Patten#matcher method returns new Matcher instance.
So, All requests does not share the state.

The purpose of this change is to reduce the compile cost of the Pattern.

perwendel:
@t-kameyama 
Yep, you're correct. It's the matcher that shouldn't be shared between threads but we're not doing that!

PR #501, opened 2016-03-28
Update startup listening log line for Spark

jylin:
Having the protocol in the string allows a clickable link in IDE's like IntelliJ.


perwendel:
This will show incorrect info if Spark is configured for HTTPS. secure(keyStore...

PR #510, opened 2016-04-02
Add the ability to expose the underlying server's JMX counters

gcuni:
This is extremely useful for monitoring servers. The setting is off by default so no behavior change for existing users. To turn it on simply invoke enableServerJMX(true) in the Spark class.

Augmented unit tests to do a sanity check on this as well.


bruno-ortizgb:
Any news on when this is going to be merged?

gcuni:
@perwendel - you guys interested in this? I've noticed several people in the google groups asking about monitoring of spark java in production. Being able to turn on JMX on jetty and have a tool like datadog automatically forward jmx counters to their service is quite valuable on that front.

tipsy:
@gcuni: @perwendel is on an offline vacation for a few weeks. I see no reason why we wouldn't be interested in this though.

tipsy:
@gcuni Are you still interested in merging this? Should I review the code?

gcuni:
Yes, still very interested! Please review and let me know.

src/main/java/spark/Service.java
```diff
@@ -130,6 +131,16 @@ public synchronized Service port(int port) {
     }
 
     /**
+     * Set whether to enable the underlying server's JMX counters
+     * @param enableServerJMX
+     * @return
+     */
+    public synchronized Service enableServerJMX(boolean enableServerJMX) {
```
tipsy: Does it make sense to disable it again programatically? I imagine you either want it on or off.
It should be enough to make this disabled by default and give users the option to enable it. 
`enableServerJmx()`(no args).


Is having this running resource intensive? If not it could be enabled by default 🤔 

You probably can't do this after the server has started, so add

```
if (initialized) {
    throwBeforeRouteMappingException();
}
```

src/main/java/spark/embeddedserver/jetty/JettyServer.java
```diff
@@ -44,6 +48,11 @@ public static Server create(int maxThreads, int minThreads, int threadTimeoutMil
         } else {
             server = new Server();
         }
+        if(enableServerJMX) {
```
tipsy: Formatting (there is an intellij-formatter in the repo)

src/main/java/spark/Spark.java
```diff
@@ -1010,6 +1010,9 @@ public static void webSocket(String path, Class<?> handler) {
         getInstance().webSocket(path, handler);
     }
 
+    public static void serverJMXEnable(boolean enabled) {
```
tipsy: Signature should be identical to the one in `Service`, so same name an no args.

src/main/java/spark/Service.java
```diff
@@ -64,6 +64,7 @@
     protected int maxThreads = -1;
     protected int minThreads = -1;
     protected int threadIdleTimeoutMillis = -1;
+    protected boolean enableServerJMX = false;
```
tipsy: serverJmxEnabled ? (with short comment describing what it is)

src/main/java/spark/Service.java
gcuni: Makes sense - will do. I don't have numbers comparing the two as the system in use previously had JMX stats enabled so we had to have them on in new code from the beginning to maintain monitoring parity.

tipsy:
Looks good @gcuni. Could you also update to fix the conflicting files?

gcuni:
Updated and fixed conflicts.

gcuni:
With the changes in 2.6.0 I don't think this PR is needed. I configured JMX using the new functionality and it worked, so I'm closing this.

bruno-ortizgb:
@gcuni can we have an example of how did u do that? thx

gcuni:
Sure. This is what we're doing. I've stripped out some of our code that deals with our way of looking up properties. Replace that with whatever system you're using. You will need to add the jetty jmx jar to your maven/gradle dependencies.
```
final int maxThreads = 50; //get this from env/property lookup system
final int minThreads = 8; //get this from env/property lookup system
final int idleTimeout = 60000; //get this from env/property lookup system
final int maxQueueCapacity = 1000; //get this from env/property lookup system
        
EmbeddedServers.add(EmbeddedServers.defaultIdentifier(), new EmbeddedJettyFactory((maxT, minT, threadTimeoutMillis) -> {
            Server server;

            if (maxT > 0) {
                int max = (maxT > 0) ? maxT : 200;
                int min = (minT > 0) ? minT : 8;
                int idleTimeout1 = (threadTimeoutMillis > 0) ? threadTimeoutMillis : 60000;
                if(maxQueueCapacity > 0) {
                    BlockingArrayQueue<Runnable> queue = new BlockingArrayQueue<>(maxQueueCapacity);
                    server = new Server(new QueuedThreadPool(max, min, idleTimeout1, queue));
                } else {
                    server = new Server(new QueuedThreadPool(max, min, idleTimeout1));
                }
            } else {
                server = new Server();
            }

            MBeanContainer mbContainer = new MBeanContainer(ManagementFactory.getPlatformMBeanServer());
            server.addEventListener(mbContainer);
            server.addBean(mbContainer);
            return server;
        }));

//call Spark with the supported 3 settings. Queue size is not here but our custom built server gets that from its scope
Spark.threadPool(maxThreads, minThreads, idleTimeout);
```

PR #511, opened 2016-04-04
Add method to set response content length

rubenochiavone:
This method is a convenient way to set response content length.

It is equivalent to do `Response#header("Content-Length", String.valueOf(length))`, but it uses `javax.servlet.ServletResponse` API and automatically set the length on `spark.Response#body(String)` if not already set.

### DIFF
--- a/src/main/java/spark/Response.java
+++ b/src/main/java/spark/Response.java
@@ -17,6 +17,7 @@
 package spark;
 
 import java.io.IOException;
+import java.nio.charset.Charset;
 
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletResponse;
@@ -38,6 +39,7 @@ public class Response {
 
     private HttpServletResponse response;
     private String body;
+    private boolean contentLengthManuallySet = false;
 
     protected Response() {
         // Used by wrapper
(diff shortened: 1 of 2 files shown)


src/main/java/spark/Response.java
```diff
@@ -67,12 +69,37 @@ public void type(String contentType) {
...
+                    : response.getCharacterEncoding();
+
+            Charset charset = null; //Charset.forName(charsetValue);
+
+            if (charset != null) {
+                response.setContentLength(this.body.getBytes(charset).length);
+            } else if (LOG.isDebugEnabled()) {
+                LOG.debug("unable to retrieve charset for name '{}'", charsetValue);
+            }
+        }
```
dinesh707: i would refractor this method to 

if (!contentLengthManuallySet) {
  response.setContentLength(calculateContentLnegth(this.body));
}

and create a new method with "calculateContentLnegth".

src/main/java/spark/Response.java
```diff
@@ -67,12 +69,37 @@ public void type(String contentType) {
...
+        if (!contentLengthManuallySet) {
+            String charsetValue = response.getCharacterEncoding() == null ?
+                    "ISO-8859-1" // recommended charset according to http://www.ietf.org/rfc/rfc2047.txt
+                    : response.getCharacterEncoding();
+
+            Charset charset = null; //Charset.forName(charsetValue);
+
+            if (charset != null) {
+                response.setContentLength(this.body.getBytes(charset).length);
+            } else if (LOG.isDebugEnabled()) {
```
dinesh707: im not 100% sure abut this?, isnt those new loggers do this already. Do you have to write a if condition to check debug ?. Any way your string is passed with {}. so unless the logger get activated it should not do string operations

src/main/java/spark/Response.java
rubenochiavone: I agree. It would be much better.

src/main/java/spark/Response.java
rubenochiavone: I dont know. I coded it using the code style of the class.

rubenochiavone:
I noticed that I left a bug: `Charset charset = null; //Charset.forName(charsetValue);` that should only be `Charset charset = Charset.forName(charsetValue);`. Fix it and also added the `calculateContentLength()` recommended method.

tipsy:
Could you very briefly explain what the use case for this is?

rubenochiavone:
The use case is when you need to explicitly set the response content length header and do not want to manually calculate the content length.

As I mentioned it is just a convenient method because something like `Response#header("Content-Length", String.valueOf(length))` already does it.

tipsy:
@rubenochiavone Sorry it took so long for a reply here. I appreciate the effort, but I think this is a little too much code for the functionality it provides.

Issue #503, opened 2016-03-29
Get response status code


posto:
Please provide a method to allow getting the status code of an http response.
The use-case is to log something in post-request filters (.after), based on response code.

tipsy:
You can use ```response.raw().getStatus()``` for now.

tipsy:
Fixed in #512

PR #512, opened 2016-04-04
Fix for #503

t-kameyama:
add `Response#status()` and `Response#type()`


perwendel:
Thanks for contributing!

PR #515, opened 2016-04-05
Fix for #251

t-kameyama:
After invalidating a session, returns new session.
```
Session session = request.session();
session.invalidate();
session = request.session(); // get new sesssion
session.attribute("XXX", "YYY"); // don't throw exception
```

Add the `validSession` to conditions to get the new session in Request.


perwendel:
tx

Issue #513, opened 2016-04-04
Routes, TemplateEngine and ModelAndView [discussion]


tipsy:
The current syntax for using template engines is:

``` java
get(path, (request, response) -> {
    return new ModelAndView(model, templatePath);
}, new VelocityTemplateEngine());
```

If you have a class with all your route declarations, it will look something like this:

``` java
get(path1, Class::methodReturningModelAndView1, templateEngine);
get(path2, Class::methodReturningModelAndView2, templateEngine);
get(path3, Class::methodReturningModelAndView3, templateEngine);
```

I don't like this, so what I usually do is: 

``` java
get(path, (request, response) -> {
    return render(model, templatePath); //static method
});
```

Which gives you:

``` java
get(path1, Class::methodReturningString1);
get(path2, Class::methodReturningString2);
get(path3, Class::methodReturningString3);
```

I like this a lot more. In my mind a GET shouldn't be concerned with models, views, or template engines. It should map a path to some method that produces output.

I think we should deprecate the `get(path, modelAndView, templateEngine)` method.

Pros: 
- No overloading (one way of doing things)
- Cleaner code, clearer mapping between path/ouput
- Fewer concepts for new users to familiarize themselves with

Cons:
- ?

Thoughts? Am I missing something here?

bademux:
Please take a look into #483 - we already have bicycle to ride:

Function<Result, String> render = result -> doExtraRenderingThere(result);
handler.andThen(render);

this way we can exdend `java.util.function.BiFunction<Request, Response, Object>` to do pre\post processing in flexible way with KISS interface from spark.

Spark side API:

```java
class Route{
 static class Builder{
   static Buider build(HttpMethod method, String path, BiFunction<Request,Response, String> handler){...}
   static Buider buildGet(String path, BiFunction<Request,Response, String> handler){...}

   Builder accept(MimeType type){...}

  Route toBuilder(){}
 }
}

class Spark{
 static class Builder{
  Builder addRoute(Route route);
...
  void build(){
  //process all routes here
  }
 }
}
```

carsongross:
Hi There,

I agree that the current template syntax is a bit awkward.  Typically people are going to select one template engine and go with it, so including it in every route is redundant.  Also, currently you have to pick between returning a string or using a template, and there are cases where that isn't ideal.  Finally, I think we could improve the api so that people don't need to know about models and views, and can just invoke the template directly.

I've created a proposal for a new template syntax here:

[...]

PR #516, opened 2016-04-06
Add routes/filters that take just request, just response, and neither:

tipsy:
- RequestRoute: handle(request)
- ResponseRoute: handle(response)
- EmptyRoute: handle()
- RequestFilter: handle(request)
- ResponseFilter: handle(response)
- EmptyFilter: handle()

Also, overload get/post/before/after with Request/Empty Route/Filter for lambdas

### DIFF
--- a/src/main/java/spark/EmptyFilter.java
+++ b/src/main/java/spark/EmptyFilter.java
@@ -1,0 +1,15 @@
+package spark;
+
+/**
+ * @author David Åse
+ */
+@FunctionalInterface
+public interface EmptyFilter extends Filter {
+
+    default void handle(Request request, Response response) throws Exception {
+        handle();
+    }
+
+    void handle() throws Exception;
+
+}
--- a/src/main/java/spark/EmptyRoute.java
+++ b/src/main/java/spark/EmptyRoute.java
@@ -1,0 +1,15 @@
+package spark;
+
+/**
+ * @author David Åse
+ */
+@FunctionalInterface
+public interface EmptyRoute extends Route {
+
+    default Object handle(Request request, Response response) throws Exception {
+        return handle();
+    }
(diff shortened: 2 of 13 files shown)


bademux:
Just curious, why to not let OOP to do the job? 
Just leave generic Route\Filter methods like `public static void get(String path, Route route)`
Imho, the Spark is greate cos it simple there is no need in complication by adding RequestRoute\RequestFilter it can be done on user side if needed.

tipsy:
The commit mostly just adds some simple boilerplate and overloads. Are you worried that having these extra interfaces will confuse users? 

Being able to do:

```
get("/hello", () -> "Hello");
get("/hello", req -> "Hello");

EmptyRroute route = () -> {
   // Whatever is returned is independent of request/response
}

RequestRoute route = (Request req) -> {
   // Whatever is returned depends only on request
}
```

Is good for conveying the intent of the programmer, and it makes it a lot easier to read for someone who is unfamiliar with the code.

It's hard to draw a line between what users should implement themselves and what should be in the framework, but in this case I don't see how anyone would implement it themselves without adding a big chunk of non-project code to their projects.

bademux:
Thanks for the answer!
In my opinion KISS for Spark is the 1st target - we already had Spring.

Referring to #513, `get("/hello", () -> "Hello")` can be implemented with standard `Supplier<T>` and additional method in builder.

```java
class Route{
 static class Builder{
   ...
   static Buider buildGet(String path, Supplier<String> handler){
     handler = (req, resp) -> handler.get();
   }

   Builder accept(MimeType type){...}

   //allows to "clone" similar routes
   Route toBuilder(){}
 }
}
```

tipsy:
Will not be added to core.

Issue #413, opened 2015-12-09
Not possible to override headers


tipsy:
The **header(String header, String value)** method only adds headers (contrary to what the javadoc states)

```
response.header("X-Frame-Options", "DENY");
response.header("X-Frame-Options", "SAMEORIGIN");
//X-Frame-Options is now "DENY, SAMEORIGIN"
```

I suggest setHeader and addHeader.

**Current workaround**
`response.raw().setHeader("X-Frame-Options", "SAMEORIGIN");`

qwIvan:
Can you remove it before add?

tipsy:
You can't remove a header. Not in Spark, and not in HttpServletResponse (the class Spark's Response wraps).

The implementation of **header(...)** should be changed from using **httpServletResponse.addHeader(...)** to using **httpServletResponse.setHeader(...)**, as adding multiple values to the same header is probably a lot less common than just setting them. If you really want to add multiple values you could always do **header("Header", header("Header") + "New value")**

PR #526, opened 2016-04-26
Allow overwriting headers

nihalp:
Addresses #413 
- Added ability to either set or add header
- Default behavior is to set a header (since this is likely more common)
- Modified tests to verify correct methods get called


tipsy:
Thanks for your work on this, but we've decided we don't want to change how the API works due to backwards compatibility. We'll fix this in version 3.0.

PR #532, opened 2016-04-28
Make Spark extendable

tipsy:


joeyb:
This would be super useful. As-is you have to use something like PowerMock for tests if you need to mock out `Service` since it is final.

perwendel:
Have to contemplate this one

PR #548, opened 2016-05-08
Added support for reading resources from multiple locations in classpath

### LINKED ISSUE #547: Support reading static files from multiple sources in classpath (jars and/or physical folders)
The current implementation of classpath static resource handling does not allow in some cases to read static files splitted in both physical folders and jars. The behaviour is spark is dependant of the classpath entries order.

This is problematic when deploying on tomcat because the war is unpacked so if the war contains some static files, all static files in jar dependencies are not accessible. It's possible to disable war unpacking but the webapp performance is affected.

The issue can be hard to apprehend ; I can provide more details with examples if needed.

I've written a patch which fixes the issue in PR #548 .

MouettE-SC:
- Changed ClassPathResourceHandler to generically read classpath contents, whether files are physical or inside jars
- Removed JarResourceHandler
- Slightly changed the ClassPathResourceHandlerTest tests to support the code changed

Overall the functionality is not modified. This PR fixes #547


tipsy:
Looks great, thanks!

MouettE-SC:
Thank you. There's however something that bugs me : With the following layout in the static file folder :

/public/welcome/index.html

* If the folder welcome is in a physical path in the classpath:
** GET /welcome will serve the index.html file,
** GET /welcome/ will also serve the index.html file.
* If the folder welcome is inside a jar in the classpath:
** GET /welcome will raise a 404 error,
** GET /welcome/ will serve the index.html file.

The behaviour is not consistent between jars and physical files. It was already done this way in 2.5 so I left it as-is. Do you think we should left it this way ?

tipsy:
I think it should work the same for jar and physical. It was probably just something Per missed when rushing to fix the other bug. What do you think @perwendel?

perwendel:
yep, it should work the same.

MouettE-SC:
OK them I'll make the necessary changes to get it work the same.

perwendel:
:+1:

MouettE-SC:
Sorry just noticed the code does not respect spark formatting. I'll fix it.

MouettE-SC:
All done. The behaviour is consistent when accessing static resources in the classpath. A leading '/' is always needed to directly open a directory (and its welcome file).

Note that the behaviour for external static location (ie not in classpath) remains unchanged, people can use directory paths without leading '/' to access the welcome file.

MouettE-SC:
Updated PR to the current tree in case merging should be considered.

tipsy:
Thanks, I thought this was merged already. I'll have a look tomorrow and merge it unless @perwendel objects.

perwendel:
Checking this now

perwendel:
Thanks for contributing!

PR #551, opened 2016-05-10
Guess MIME-type for static files

tipsy:


MouettE-SC:
Wouldn't it be a good idea to use a mime.property file containing the mapping instead of hard-coding the values ? There's no resource files in the project so maybe it's conventional ...

tipsy:
> Wouldn't it be a good idea to use a mime.property file containing the mapping instead of hard-coding the values ? There's no resource files in the project so maybe it's conventional ...

My reasoning was that this would be faster, simpler, and more self contained this way. If we create a property file we need to read it and populate a map manually, and it wouldn't really be any less "hard-coded"? Am I missing something?

kshep92:
I say store the mime types in a list and provide a method for users to register a new mime type when the app starts. Somerhing like Spark.registerMimeType(<mime type>, <extension>)

tipsy:
@kshep92 I'm not sure about that solution. Users shouldn't have to care about mime-types at all. We could have a registerMimeType method for adding additional mime-types, but we have to be able to guess the most common ones out of the box.

kshep92:
@tipsy Yes, my suggestion was for registering new mime types.

nikku:
Just for reference, the [mime types Nginx knows about](https://www.nginx.com/resources/wiki/start/topics/examples/full/#mime-types), out of the box. I'd rather have an ability to extend that mapping with my own mime types (if needed), rather than including an endless list.

tipsy:
Created #589 for a cleaner history.

PR #555, opened 2016-05-12
Added method signatures for multiple filters; Allow custom 404 and 500 pages

kshep92:
- Adds method signatures to allow multiple filters to be applied to before/after actions.
- Allows you to specify custom 404 and 500 error pages by placing a 404.html and a 500.html file in the resources folder. Falls back to default behaviour if files are not present.

### DIFF
--- a/src/main/java/spark/Spark.java
+++ b/src/main/java/spark/Spark.java
@@ -164,6 +164,18 @@ public class Spark {
     }
 
     /**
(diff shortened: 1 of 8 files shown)


AzureusNation:
I don't think its a good idea to load a 404 or 500 page every request that needs it. Shouldn't this be added to spark ignite and then grabbed in the route handler?

tipsy:
Custom error pages were added in #719

Issue #448, opened 2016-01-19
Difficulty  to read dynamic path parameters


jamsheer-thottathil:
I have a end point and  tail contains with its return type's extension

Example :name/test.json

Here i can read path parameter as  follows  

```
get("/:name/test.json", (request, response) -> {
    return "Hello: " + request.params(":name");
});
```

But i am getting difficulty to read path parameters if  it is in last part of URL 

Example test/:name.json

Here i need to read  as request.params(":name")  ,Is it possible with spark .

Jersey , restlet  like frameworks are  support it by the help of {}.

craigsdennis:
With the current implementation, this doesn't look to be possible.

The entire part of the path is stored, see the [Request code here that gets the parameters](
https://github.com/perwendel/spark/blob/master/src/main/java/spark/Request.java#L462)

Unfortunately, in your example, `request.params(":name.json")` will return the entire portion and match anything in that part of the path.

perwendel:
I have difficulties seeing how this an actual problem. Why not just read the parameter value and parse the string to the two parts you need? E.g. with split(".") ?

jamsheer-thottathil:
Than you all for your reply  and I am sorry ,i did not get you @perwendel 

In my example i cannot read name value as its is available in URL with .json tail.

Here 'name' is my variable name and .json came apart  of my rest standard

tipsy:
Added to v3 roadmap

adityakapre2009:
Is v3 released yet ?

PR #561, opened 2016-05-18
 #448 Implementation to read dynamic path parameters

alissonbrunosa:
Example usage

``` java
import static spark.Spark.*;

public class HelloWorld {
        get("/hello/:name.json", (request, response) -> {
            return "Hello: " + request.params(":name");
        });
    }
}
```

### DIFF
--- a/src/main/java/spark/Request.java
+++ b/src/main/java/spark/Request.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


tipsy:
@alissonbrunosa could you elaborate/specify what this solves (and changes)?

alissonbrunosa:
Hi @tipsy 
If you check the issue #448, you can see what this PR solves.
The main idea is to allow the user to specify and read dynamic path.
You can check the example in the PR as well.

tipsy:
@alissonbrunosa Dynamic path params are already support via `:param`. This PR seems to allow adding file-extensions to path-params, but in which cases is that useful?

If you just do `request.params("name.json")` you will get the value you need, and if you do `request.params("name.json").split(".json")[0]` you will replicate what your PR does, correct?

I'm not seeing a big need for this, and I'm worried that it might cause unexpected behavior 🤔

alissonbrunosa:
@tipsy You're completely correct.
I've just added this feature because **Spark** is inspired by Sinatra. If you take a look at Sinatra, you'll see it has this feature.

tipsy:
Ah, I'll all it do the v3 then!

PR #572, opened 2016-05-30
added missing delegate implementation in ResponseWrapper

Born2Go:
this missing delegate causes a null pointer exception when using "public void cookie(String path, String name, String value, int maxAge, boolean secured, boolean httpOnly) "


anttipoi:
I just run into a similar issue: methods 
`public int status()`
and 
`public String type()`
are missing overrides in ResponseWrapper, too. 

I tried accessing response.status() in my after-wrapper and ran into a NPE.

Maybe these should be overridden as well?

PR #574, opened 2016-06-02
Exclude cipher suites and protocols in a secure connector

vzamanillo:
Added the hability to create a secure socket connector removing a desired cipher suite or protocol list via 

`sslContextFactory.addExcludeProtocols(String[]);`

and

`sslContextFactory.addExcludeCipherSuites(String[]);`

Example of how to use it:

``` java
package com.my.test.spark;

import static spark.Spark.get;
import static spark.Spark.secure;

public class HelloWorld {
        secure("/tmp/keystore.jks", "kspasswd", "/tmp/trustore.jks", "tspasswd", null, new String[] { "TLS_DHE_RSA_WITH_AES_128_CBC_SHA256",
                "TLS_DHE_RSA_WITH_AES_128_CBC_SHA", "TLS_DHE_RSA_WITH_AES_128_GCM_SHA256", "TLS_DHE_RSA_WITH_3DES_EDE_CBC_SHA" });
        get("/hello", (req, res) -> "Hello World");
    }
}
```

Results tested with SSL Scan

```
Testing SSL server localhost on port 4567

  Supported Server Cipher(s):

Rejected  TLSv1  128 bits  DHE-RSA-AES128-SHA
```

results using the default secure method

``` java
package com.my.test.spark;

import static spark.Spark.get;
import static spark.Spark.secure;

public class HelloWorld {
        secure("/tmp/keystore.jks", "kspasswd", "/tmp/trustore.jks", "tspasswd");
        get("/hello", (req, res) -> "Hello World");
    }
}
```

```
Testing SSL server localhost on port 4567

  Supported Server Cipher(s):

Accepted  TLSv1  128 bits  DHE-RSA-AES128-SHA
```


src/main/java/spark/resource/AbstractResource.java
```diff
@@ -120,7 +120,7 @@ public File getFile() throws IOException {
     @Override
     public long contentLength() throws IOException {
         InputStream is = this.getInputStream();
-        Assert.state(is != null, "resource input stream must not be null");
+        Assert.isTrue(is == null, "resource input stream must not be null");
```
nkavian: The logic and the comment don't match.  If this.getInputStream() returns null, the equality will be true, and the assert will pass.  But the comment says input stream must NOT be null.

src/main/java/spark/utils/MimeParse.java
```diff
@@ -116,12 +116,10 @@ public int compareTo(FitnessAndQuality o) {
             if (fitness == o.fitness) {
                 if (quality == o.quality) {
                     return 0;
-                } else {
-                    return quality < o.quality ? -1 : 1;
                 }
-            } else {
-                return fitness < o.fitness ? -1 : 1;
+				return quality < o.quality ? -1 : 1;
```
nkavian: Shouldn't this line be moved one line higher, before the close brace?

src/main/java/spark/utils/MimeParse.java
```diff
@@ -116,12 +116,10 @@ public int compareTo(FitnessAndQuality o) {
...
             }
+			return fitness < o.fitness ? -1 : 1;
```
nkavian: Shouldn't this line be moved one line higher, before the close brace?

src/test/java/spark/examples/books/Books.java
```diff
@@ -58,10 +58,9 @@ public static void main(String[] args) {
             Book book = books.get(request.params(":id"));
             if (book != null) {
                 return "Title: " + book.getTitle() + ", Author: " + book.getAuthor();
-            } else {
-                response.status(404); // 404 Not found
-                return "Book not found";
             }
+			response.status(404); // 404 Not found
```
nkavian: You're indention is off in various places.

nkavian: Also, as example code, the if/else structure is more readable in that context.

src/main/java/spark/Service.java
```diff
@@ -162,6 +162,26 @@ public synchronized Service secure(String keystoreFile,
         sslStores = SslStores.create(keystoreFile, keystorePassword, truststoreFile, truststorePassword);
         return this;
     }
+ 
+    public synchronized Service secure(String keystoreFile,
+							           String keystorePassword,
+							           String truststoreFile,
+							           String truststorePassword,
+							           String[] excludeProtocols,
```
nkavian: Traditionally, when configuring nginx and Apache web servers, they work with include lists because it's treated as a whitelist.  Using a blacklist/exclude list has the downside that new protocols and ciphers automatically become enabled.

Checkout: http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/util/ssl/SslContextFactory.html#setIncludeCipherSuites-java.lang.String...-
void setIncludeCipherSuites(String... cipherSuites)
void	setIncludeProtocols(String... protocols)

Using an include list also makes it much easier for an Ops team to work with SparkJava.

src/main/java/spark/resource/AbstractResource.java
bellzebu: You are completly right, sorry.

src/main/java/spark/utils/MimeParse.java
bellzebu: No, it's right.

src/main/java/spark/utils/MimeParse.java
bellzebu: No, it's right.

src/test/java/spark/examples/books/Books.java
bellzebu: You are right, is more readable but is not necessary, it should be more readable using a method variable and one return statement only but is not the case, anyway this is about coding styles, maybe should not be accepted.

src/main/java/spark/Service.java
bellzebu: That's the point, the embedded jetty server of Spark offers the default cipher suites and protocols of the JVM with 0 config, so we must assume that ciphers are secure and not deprecated. The applied policy (blacklist) is the same policy of the `org.eclipse.jetty.util.ssl.SslContextFactory`, _"run jetty with the dedault cipher suites of the JVM and blacklist the deprecated or insecure cihpers or protocols"_, an example

```java
    /**
     * Construct an instance of SslContextFactory
     * Default constructor for use in XmlConfiguration files
     * @param trustAll whether to blindly trust all certificates
     * @see #setTrustAll(boolean)
     */
    public SslContextFactory(boolean trustAll)
    {
        setTrustAll(trustAll);
        addExcludeProtocols("SSL", "SSLv2", "SSLv2Hello", "SSLv3");
        setExcludeCipherSuites(
                "SSL_RSA_WITH_DES_CBC_SHA",
                "SSL_DHE_RSA_WITH_DES_CBC_SHA",
                "SSL_DHE_DSS_WITH_DES_CBC_SHA",
                "SSL_RSA_EXPORT_WITH_RC4_40_MD5",
                "SSL_RSA_EXPORT_WITH_DES40_CBC_SHA",
                "SSL_DHE_RSA_EXPORT_WITH_DES40_CBC_SHA",
                "SSL_DHE_DSS_EXPORT_WITH_DES40_CBC_SHA");
}
```

I think is more easy to do blacklist for one or two insecure ciphers than doing a whitelist and including (again) all the default and secure ciphers from JVM.

tipsy:
@vzamanillo Do you still want this merged? If yes, can you remove all the un-related fixes?

vzamanillo:
@perwendel , Hi, ok I will remove the un-related fixes ASAP, thnx.

vzamanillo:
@perwendel, all un-related fixes were removed.

Issue #570, opened 2016-05-28
Static instance of ExceptionMapper.


nkavian:
When using the new multi-service feature, the wrong exception mapper is used between 2 services because it's using a static shared exception mapper: https://github.com/perwendel/spark/blob/114d245ef68b1f25d9793d099ae1ca25bd63f1b7/src/main/java/spark/Service.java#L390

melges-morgen:
Guys, can i hope that this bug will be fixed, as far as I can see, pool-request is ready?

perwendel:
Fixed by #1033

PR #576, opened 2016-06-06
#570 Fix: Static instance of ExceptionMapper. 

vzamanillo:
Per service ExceptionMapper.


jakaarl:
I haven't really looked into the reasoning/logic of this, but on a superficial level, this looks good to me. So the idea is, that it's per `Servlet` or per `Service`, depending on if being run in embedded or deployed mode? @vzamanillo could you resolve the conflicts?

vzamanillo:
@jakaarl, I don't remember the applied logic actually, the PR is very old and I need to be review it. The merge conflicts are resolved.

melges-morgen:
Hi, does anybody looking on it? I am faced with bug which should be fixed by this PR and it would be great if it will be merged.

vzamanillo:
The PR is ready to be merged.

PR #579, opened 2016-06-12
add missing delegate methods to ResponseWrapper

anttipoi:
As I commented in PR [572](https://github.com/perwendel/spark/pull/572), there are other methods missing from the delegation, too.

I noticed the omission when writing an after-filter which tried to access response.status() and got an NPE. 

This PR adds the methods (and a test demonstrating the problem).


zmalik:
Nice! this PR also fixes the issue #592

dherges:
see also #600 and #599 .. all duplicates

Issue #446, opened 2016-01-08
A way to get the port if you specify `port(0)`


dkowis:
As best as I can tell, once you do this, there's no way to get the port programmatically. I haven't found a way to get to the underlying Jetty server to get those port details, and the port parameter can only be set, never queried.

It would certainly make integration style testing a whole lot easier, and more reliable, if I could start spark on a random port, and then inspect that port for use in my integration tests.

dkowis:
For cross linking, in case someone comes across it here, I have done dirty reflection hacks to get at this port.
http://stackoverflow.com/a/34687059/423218

I would love for there to be an API approved way to get the port for use during integration testing.

IRus:
Related to #400

PR #588, opened 2016-06-17
Issue-446 Allow a user to retrieve the allocated port the server is r…

chibenwa:
…unning on

When writing tests, I don't want my test to fail if the port I want to run Spark on is used by an other program.
Thus, I need to use a random port for Spark.
However, I have no mean to retrieve it from my code, making the server impossible to test.

Related to https://github.com/perwendel/spark/issues/446 and https://github.com/perwendel/spark/issues/400

The basic idea behind this PR is that the Embedded server should return the port it was effectively launched on. Hence we can retrieve the allocated port without changing the logic.

### DIFF
--- a/src/main/java/spark/embeddedserver/EmbeddedServer.java
+++ b/src/main/java/spark/embeddedserver/EmbeddedServer.java
@@ -38,8 +38,10 @@ public interface EmbeddedServer {
      * @param maxThreads              - max nbr of threads.
      * @param minThreads              - min nbr of threads.
      * @param threadIdleTimeoutMillis - idle timeout (ms).
+     *
+     * @return The port number the server was launched on.
      */
-    void ignite(String host,
+    int ignite(String host,
                 int port,
                 SslStores sslStores,
                 CountDownLatch latch,
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -132,6 +132,21 @@ public final class Service extends Routable {
     }
 
     /**
+     * Retrieves the port that Spark is listening on.
+     *
+     * @throws IllegalStateException when the server is not started
+     *
+     * @return The port Spark server is listening on.
+     */
+    public synchronized int getPort() {
+        if (initialized) {
+            return port;
+        } else {
+            throw new IllegalStateException("This must be done after route mapping has begun");
+        }
+    }
+
+    /**
      * Set the connection to be secure, using the specified keystore and
      * truststore. This has to be called before any route mapping is done. You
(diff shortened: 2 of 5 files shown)


qmg-gkalogiros:
any ideas regarding when this is going to be merged?

AzureusNation:
There is one huge flaw with this PR. You still need to assign a port number or its going to fail running 2 instances of Spark. All this has really added is a couple of methods to retrieve the port number. Personally I don't see why this is needed?

src/main/java/spark/embeddedserver/EmbeddedServer.java
```diff
@@ -38,8 +38,10 @@
      * @param maxThreads              - max nbr of threads.
      * @param minThreads              - min nbr of threads.
      * @param threadIdleTimeoutMillis - idle timeout (ms).
+     *
+     * @return The port number the server was launched on.
      */
-    void ignite(String host,
+    int ignite(String host,
```
ruurd: Why? You added the accessor for the port, that should do it right?

ruurd:
If you don't want your tests to fail because the SUT can't allocate the port to listen on, then test if it can be allocated before you start testing.

src/main/java/spark/embeddedserver/EmbeddedServer.java
AzureusNation: This needs to return the port number or else the Service object wont know that the port number has been changed and will keep returning the default port (4567).

AzureusNation:
I take back my other comment. (It was late for me). I agree fully with this PR and should be merged.

src/main/java/spark/embeddedserver/EmbeddedServer.java
ruurd: Ah. OK I missed that.

chibenwa:
@ruurd : I do not understand your last comment. How can I modify it to match your needs ?

ruurd:
@chibenwa in other words - before you start testing see if the port in question is allocated or not. If not, test, if taken, fix first. In general testing should be done in a known environment.

jakaarl:
While I in principle agree with @ruurd here, I think "an available port assigned by the underlying implementation" is known enough, and most of all, convenient.

perwendel:
I changed from `getPort()` to `port()` before merging!

lyonzy:
Anyone had any luck using this? It seems the port is never returned back to Service from ignite() because `server.join();` is waiting for all threads to finish...

jakaarl:
@lyonzy is right. See #714.

PR #594, opened 2016-06-24
Support for multiple Spark applications in filter configuration

sock-osg:
I have been working with the API and i think that it is great, i like keeping my code clean as much as possible and i think that a way to achieve it is that the end-points could be grouped into business modules. That's why i made some modifications in order to be able to support the configuration of multiple spark applications.


perwendel:
Good stuff!

PR #596, opened 2016-06-26
Add web fonts mime types to MimeType inner map

giflw:
Hi,

As Spark will be guessing some mime types, would be nice to have web fonts mime types too.

Best regards.


tipsy:
Good idea. Coul you add .woff2 while you're at it?

giflw:
Added woff2 mime type =D

tipsy:
Thanks!

Issue #248, opened 2015-03-02
Embedded Jetty session leak


phront:
We found that under heavy load our spark instance had suffered a session leak.

*Class    class org.eclipse.jetty.server.session.HashedSession*
*Instance Count   1603602*
*Total Size           145 927 782*

It seems that Jetty's *SessionManager* does not limit *maxInactiveInterval* by default so an "out of box" Spark should set it manually to any reasonable value.

PR #603, opened 2016-07-02
Fix for Embedded Jetty session leak #248

AzureusNation:
This fixes the leak of HashedSession instances in the builtin Jetty Web Application Server.


tipsy:
Should also be fixed by #813

PR #604, opened 2016-07-02
Logging error when not caught by an exception handler

izzabudaka:
Logging error when not caught by an exception handler.

I had a problem when I was working with spark and did not have an ExceptionHandler mapped. The exception that was thrown was not logged. This led to a very lengthy debugging session which could have avoided by logging the exception.


MouettE-SC:
Duplicate of #487

Issue #373, opened 2015-10-23
Internal Static Files are served as "text/plain" when using tomcat


Entwanderer:
I'm using spark with tomcat8. 
```javascript
staticFileLocation("/public")
```
seemed to work so far.

But there is a problem, all the files are served with content type "text/plain".
So when including CSS files with absolute path
```javascript
<link rel="stylesheet" href="/example.css"/>
```

it will work and be used in FireFox and Chrome. But Chrome throws an error that it must be "text/css". And Internet-Explorer (even the newest Edge-Version) does also so, and is even not loading the CSS!

See SEC7113: "SEC7113 - CSS was ignored due to mime type mismatch"
https://msdn.microsoft.com/en-us/library/hh180764%28v=vs.85%29.aspx

Entwanderer:
I found a way to solve this, but I'm not sure if this is best practice or intented to be done this way.

I extended the class SparkFilter the following way (unoptimized code):
```javascript
public class ApplicationFilter extends SparkFilter
{
    @Override
    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException
    {
        String requestUrl = ((HttpServletRequest) request).getRequestURI().toString();

        Map<String, String> mimeMapping = new HashMap<>();
        mimeMapping.put(".css","text/css");
        mimeMapping.put(".js","text/javascript");

        for(Map.Entry<String,String> entry : mimeMapping.entrySet())
        {
            if(requestUrl.endsWith(entry.getKey()))
            {
                ((HttpServletResponse) response).setHeader("Content-Type",entry.getValue());
            }
        }
        super.doFilter(request, response, chain);
    }
}
```

And changed the class SparkFilter with the ApplicationFilter in the web.xml
```javascript
<filter-class>com.company.app.ApplicationFilter</filter-class>
```

sciops:
Thank you for this. I'd like to add that it helped me solve an issue with the root response returning text/plain instead of text/html as the content type. I used your class and added the following line:

`mimeMapping.put("/","text/html");`

clivefoley:
Thanks for this, I was pulling my hair out trying to figure out why it was working on a local tomcat but not on AWS.

MouettE-SC:
To add a little more confusing stuff, when apache is used as a reverse proxy with tomcat, it adds the content-type...

PR #618, opened 2016-07-12
Bugfix: content-type of static files

martin-pabst:
Without this commit static files are always served with content-type "application/octet-stream" as spark.staticfiles.StaticFilesConfiguration.consumeWithFileResourceHandlers calls spark.staticfiles.MimeType.fromResource to infer the content-type from the filename.
The latter calls resource.getFilename which falls back to base class implementation (which returns null) in case of ExternalResource.

### DIFF
--- a/src/main/java/spark/resource/ExternalResource.java
+++ b/src/main/java/spark/resource/ExternalResource.java
@@ -86,4 +86,14 @@ public class ExternalResource extends AbstractFileResolvingResource {
     public String getPath() {
         return file.getPath();
     }
+    
+    @Override
+    public String getFilename() {
+    	
+    	if(file != null){
+    		return file.getName();
+    	}
+    	
+    	return super.getFilename();
+    }
 }


AzureusNation:
@martin-pabst does this fix any existing issues or is this a new issue that you have found?

martin-pabst:
I opened issue #619 to describe the bug. Hopefully this bugfix also clears issue #373.

AzureusNation:
super

silvereddie:
Does anyone have ETA when this fix will be released?

MouettE-SC:
PR #589 solved it and has already been merged into the tree.

perwendel:
I'm currently on vacation without access to computer. Hence, there will be no merging of (at least complex) PR's in the 2-3 coming weeks.

src/main/java/spark/resource/ExternalResource.java
```diff
@@ -86,4 +86,14 @@ public URL getURL() throws IOException {
     public String getPath() {
         return file.getPath();
     }
+    
+    @Override
+    public String getFilename() {
+    	
+    	if(file != null){
```
perwendel: bad formatting.

acwwat:
Hi there, any update on this PR? Would really like this merged so that Chrome can display static HTML files served by Spark properly. Thanks.

perwendel:
#633 Has already been merged. This is a different solution on the same problem.

PR #621, opened 2016-07-13
Add HTTP method to route-not-mapped info

evaK-de:
The info that is logged when an incoming request uses an unmapped route only printed the path, but not the method.
That made the statement ambiguous, since you could e.g. have mapped the route for POST, but not for GET.

BEFORE:
INFO spark.http.matching.MatcherFilter - The requested route [POST /myapi/v1/blogpost/] has not been mapped in Spark

AFTER:
INFO spark.http.matching.MatcherFilter - The requested route [/myapi/v1/blogpost/] has not been mapped in Spark

### DIFF
--- a/src/main/java/spark/http/matching/MatcherFilter.java
+++ b/src/main/java/spark/http/matching/MatcherFilter.java
@@ -149,7 +149,7 @@ public class MatcherFilter implements Filter {
         }
 
         if (body.notSet() && !externalContainer) {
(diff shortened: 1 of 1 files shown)


AzureusNation:
@evaK-de Can you squash your commits please and ill poke @perwendel and @tipsy

evaK-de:
Sqaushed my changes in the next PR.

PR #623, opened 2016-07-14
Adding a method to configure max size of HTTP headers.

spuklo:
I've stumbled upon an issue that I've had to deal with client that is using huge headers (more than 16k) and Jetty's defaults were not good enough. This PR provides a Spark.maxHeadersSize(size) method that sets max headers size for both requests and responses.
This required a little change in SocketConnectorFactory class to pass HttpConnectionFactory explicitly (it has been used anyway, but previously was not created in Spark code).

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -16,14 +16,8 @@
  */
 package spark;
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.CountDownLatch;
-
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


tipsy:
@spuklo do you still want this? Could you clean up the PR and remove all the unrelated changes (the refactoring)?

spuklo:
@tipsy yes, still would love to have it in Spark. Will clean up the PR

spuklo:
@Tipsy - I'll close this PR and will re-apply changes on top of current master, then create new PR. Trying to resolve all merge conflicts between my fork and main Spark is just too much pain.

PR #633, opened 2016-08-02
Implement getFilename for ExternalResource

menski:
The [mime type detection](https://github.com/perwendel/spark/blob/12774f1f20e2871c4dd19293d08dcf5afdd484db/src/main/java/spark/staticfiles/MimeType.java#L98) is based on the filename. This commit adds an implementation of `getFilename` for `ExternalResource` so the mime type is also set on external static files.


perwendel:
tx for contributing!

PR #635, opened 2016-08-02
Implement mime type guessing for jar resources

menski:
- guess the mime type of a jar resource by the request path


menski:
I wasn't sure how to test jar resources. If someone could point me into a direction how to test this change I will add test cases for it.

nikku:
@perwendel Any update on this? Without this feature users won't be able to use the added MIME type functionality with fat jars.

tipsy:
Already implemented in some other PR.

PR #637, opened 2016-08-06
Fixed bug where null responses were being handled as not matched routes

MartinsThiago:
Here's one more

### DIFF
--- a/src/main/java/spark/http/matching/MatcherFilter.java
+++ b/src/main/java/spark/http/matching/MatcherFilter.java
@@ -148,10 +148,15 @@ public class MatcherFilter implements Filter {
             }
         }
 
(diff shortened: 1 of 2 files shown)


AzureusNation:
Looks like you've broken 11 tests. They will need fixing.

MartinsThiago:
Hmmm... After I forked the project i saw they were breaking already

It's not an excuse, but the weirdo is running these tests where spark is
ignited separately the tests passes :(

MartinsThiago:
If today I have time I'll try to investigate why it's behaving this way

MartinsThiago:
Whooops... I ran the tests through maven then I saw those error.

I ll submit a fixed PR, sorry :sweat_smile:

PR #639, opened 2016-08-07
Fixed bug where null responses were being handled as not matched routes

MartinsThiago:
Fixed previous pr -> #637

### DIFF
--- a/src/main/java/spark/http/matching/AfterFilters.java
+++ b/src/main/java/spark/http/matching/AfterFilters.java
@@ -29,13 +29,14 @@ import spark.routematch.RouteMatch;
  */
 final class AfterFilters {
 
-    static void execute(RouteContext context) throws Exception {
+    static boolean execute(RouteContext context) throws Exception {
 
         Object content = context.body().get();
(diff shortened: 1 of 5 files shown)


MartinsThiago:
`echo 'Hey o/... dont forget this one'`

tipsy:
@perwendel is on an offline vacation. He will start merging changes around the end of the month.

MartinsThiago:
Oops, sorry :)

MartinsThiago:
Got the idea

I saw some pull requests with same intention... So I vote for achieving empty responses with null even cause `""` are not trully trully empty

Sometime ago I got confused because a created a route (returning null) and I was not understanding why logs were telling me no route was matched and I saw the route being executed while debugging

MartinsThiago:
So... In summary I agree that spark should rely on `response.body` but for routes I think we should rely on `match`

perwendel:
I'll think about it :)

MartinsThiago:
Nice... I am closing this pull

Anything call me o/

PR #644, opened 2016-08-23
log which accept header is not mapped

yegeniy:
If the route is rejected because of the Accept header, it's useful to know which one was used.

Also, changing to using SLF4J's parameter `{}` syntax.

### DIFF
--- a/src/main/java/spark/http/matching/MatcherFilter.java
+++ b/src/main/java/spark/http/matching/MatcherFilter.java
@@ -149,7 +149,8 @@ public class MatcherFilter implements Filter {
         }
 
         if (body.notSet() && !externalContainer) {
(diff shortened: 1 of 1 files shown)


src/main/java/spark/http/matching/MatcherFilter.java
```diff
@@ -149,7 +149,8 @@ public void doFilter(ServletRequest servletRequest,
         }
 
         if (body.notSet() && !externalContainer) {
-            LOG.info("The requested route [" + uri + "] has not been mapped in Spark");
+            LOG.info("The requested route [{}] has not been mapped in Spark for {}: [{}]",
+                uri, ACCEPT_TYPE_REQUEST_MIME_HEADER, acceptType);
```
yegeniy: It might be clearer to only print the bit about the accept header if one was actually provided.

PR #614, opened 2016-07-10
Add mechanism of configuring the embedded web server's context path via Spark Framework

initor:
Added functionality of configuring the embedded Jetty web server's context path via Spark Framework.

PR #658, opened 2016-09-04
Expose the context path (if deployed inside a Servlet container)

mthmulders:
When a Spark application is deployed inside a Servlet container, it is oftentimes not deployed under the context root, but under a certain context instead.
This PR exposes that context path to the application. It can be used to generate e.g. links to other pages inside a Spark application, to static resources inside the application, etc.

I'm not completely sure this is the best way to do it. Feel free to suggest improvements if you feel it could be done in a better way.

### DIFF
--- a/src/main/java/spark/globalstate/ServletFlag.java
+++ b/src/main/java/spark/globalstate/ServletFlag.java
@@ -24,12 +24,15 @@ import java.util.concurrent.atomic.AtomicBoolean;
 public class ServletFlag {
 
     private static AtomicBoolean isRunningFromServlet = new AtomicBoolean(false);
+    private static String contextPath = null;
 
     /**
      * Tells the system that Spark was run from an "external" web application server.
+     * @param contextPath The context path under which the Spark application is deployed.
      */
-    public static void runFromServlet() {
+    public static void runFromServlet(String contextPath) {
         isRunningFromServlet.set(true);
+        ServletFlag.contextPath = contextPath;
     }
 
     /**
@@ -39,4 +42,10 @@ public class ServletFlag {
         return isRunningFromServlet.get();
     }
 
+    /**
+     * @return the context path under which the Spark application is deployed.
+     */
+    public static String getContextPath() {
+        return contextPath;
+    }
 }
--- a/src/main/java/spark/servlet/SparkFilter.java
+++ b/src/main/java/spark/servlet/SparkFilter.java
@@ -61,7 +61,8 @@ public class SparkFilter implements Filter {
 
     @Override
     public void init(FilterConfig filterConfig) throws ServletException {
-        ServletFlag.runFromServlet();
+        String contextPath = filterConfig.getServletContext().getContextPath();
+        ServletFlag.runFromServlet(contextPath);
 
         applications = getApplications(filterConfig);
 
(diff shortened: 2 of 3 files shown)


atomfrede:
This would be very useful when using thymeleaf!

mthmulders:
@perwendel, any chance to review and possibly merge this?

jakaarl:
The changes look OK, but I'd be a little concerned about concurrency: the servlet flag boolean is probably atomic for a reason - should the context path be wrapped in an `AtomicReference`? And more importantly, is there a hard limit of 1 `SparkFilter` per webapp? If not, this will not work correctly for several instances.

mthmulders:
Thanks for the feedback @jakaarl!

Wrapping the `contextPath` in an `AtomicReference` shouldn't be a problem. I'll take care of that.

I don't know whether you can define the `SparkFilter` multiple times inside one webapp. I'm not sure why you would want that - AFAIK you can run multiple Spark applications inside one `SparkFilter`. A quick scan through `SparkFilter` does not show me any evidence of a hard limit, so I suppose it is technically possible. Indeed, in such a case this PR will not work. Do you think it should be possible to define the `SparkFilter` multiple times inside one webapp?

jakaarl:
Sorry, my bad. I was mixing the URL mapping of the filter with the context path. Naturally, the context path is the same for all filters in the same webapp. So, shouldn't be a problem.

mthmulders:
Thanks again @jakaarl; is there anything else that prevents merging this?

tipsy:
@mthmulders did you look at #614 ?

mthmulders:
@tipsy As far as I can see, #614 handles the case where Spark is deployed as a stand-alone app with embedded Jetty. This PR, on the other hand, handles situations when a Spark app is deployed inside a Servlet container, and there is no embedded Jetty in that case. I think there is no overlap.

tipsy:
@mthmulders why are you changing the method signature of runFromServlet instead of overloading it?

mthmulders:
Since you'd always want to ensure the context path is available when deployed inside a Servlet container, it would not make sense to overload it without also setting the context path. Put differently, I think that this approach expresses the fact that running inside a Servlet container means you have to tell Spark under what context path you are deploying your app. Since I found only one use of the method (inside the `SparkFilter` that boots Spark from the Servlet container) it didn't have a lot of impact.

tipsy:
I only use Spark embedded, but I assume `runFromServlet` is a public method that's used when you run Spark inside servlet containers (how else would this commit let people set their context path?). 
If so, wouldn't this commit break backwards compatibility if you don't overload the method?

mthmulders:
According to [the documentation](http://sparkjava.com/documentation#other-webserver) (and my own experiences with Spark so far), all you need to do when you want to run in a Servlet container is configure the aforementioned `SparkFilter` in your web.xml. So although the `runFromServlet` method is public, I'm not sure you'd consider it a public API.

On the other hand, if people really feel it's an issue, it's pretty easy to make it an overload and have the original method set the `contextPath` to `"/"`. The disadvantage is that this creates room for error: people having their own `SparkFilter` might forget to invoke the overloaded method and then be faced by an incorrect `contextPath` inside Spark.  In my opinion, that would contradict the philosofy of "being as simple and straight-forward as possible" (quote from [homepage](http://sparkjava.com/)).

Issue #386, opened 2015-11-10
Internal Redirect?


haochenx:
I've used spark for a few hours, and found it extremely cool. But I cannot find something that would make my code clear and simple in the documentation: the internal redirection, also known as RequestDispatcher#forward(req, resp) in the Java Servlet world. I guess this is quite a handy feature and it should not be hard to implement. Am I missing something or is it worthy a feature request?

tipsy:
How did you picture the syntax for this being?

haochenx:
For example on the Request class, add a method like `Object Request#redispatch(String path)`, in this way, it's is possible to provide the flexibility to further modify the body.

But this opens the possibility of infinite redispatching loop, of which some mechanism should be added to assist the debugging; also special care might be needed so that this kind of loop won't chock the server.

tipsy:
Sorry, I'm still not clear on how this would work and what the purpose would be. Could you please expand on this example to demonstrate?

```java

public static void main(String[] args) {

    Route routeOne = (Request request, Response response) -> {
        return "Route One";    
    };

    Route routeTwo = (Request request, Response response) -> {
        return "Route Two";
    };
    
    get("routeOne", routeOne);
    get("routeTwo", routeTwo);
    
}
```

haochenx:
For example if you want `/routeTwo` simple being internally redirected to `/routeOne`, you could write it as below

````
public static void main(String[] args) {

    Route routeOne = (Request request, Response response) -> {
        return "Route One";    
    };

    Route routeTwo = (Request request, Response response) -> {
        return request.redispatch("routeOne");
    };

    get("routeOne", routeOne);
    get("routeTwo", routeTwo);

}
````

**Comment by tipsy** [comment]
It looks like you achieve what you want by using the  ```handle()``` method on the route: 

```java
Route routeOne = (Request request, Response response) -> {
    return "Route One";
};

Route routeTwo = (Request request, Response response) -> {
    return routeOne.handle(request, response);
};
```

haochenx:
The routeOne object may not be accessible in the routeTwo block. Also, it's quite common for a route to derive parameters based on the URL, which I think would break with such a naive invocation.

I was forced to create a route that listen "/*" and do routing logic myself.

MartinsThiago:
+1 to internal redirect

I think most people use spark for creating rest apis... 

who never went through this situation? Where you have related resources

GET /authors`
GET `

MartinsThiago:
Anyone on this?

[...]

PR #659, opened 2016-09-04
Added Internal Redirect Feature

QuackMartins:
PR about #386


MartinsThiago:
Improved redispatcher implementation, now it uses the same logic as MatchFilter

sepatel:
Any further progress on this? Is a really helpful feature that I'm looking for in order to help support backwards compatibility as we refactor the web service API's and help migrate people over to the new API's without breaking them outright.

MartinsThiago:
@sepatel I don't know if this PR is going to be merged soon, but if you really need this feature...
You can use it with [jitpack.io](https://jitpack.io/#martinsthiago/spark/2.4)

sepatel:
I'm a bit confused. The jitpack link is for version 2.4.0. The latest version is 2.6.0.RC0. Currently I'm using 2.5.5 and was planning to move over to 2.6.0 once it is properly released.

Is this something that will be part of 2.6.0 when properly released? If so then I can hold off a bit and wait for that probably as I still have another 3 weeks before this update hits production.

tipsy:
> Is this something that will be part of 2.6.0 when properly released?
No, we won't add anything new to the RC (unless it's very very small), just fix bugs.

sepatel:
So is it then a fork of the 2.4.x codebase? If so then I will probably be better of sticking with the hack I'm doing to try and redirect the responses around. What release is it scheduled for? I'm presuming then it would fall under a potential 2.7.x release? Do you have a snapshot of that upcoming release then or a fork of the 2.6.0 codebase this feature might be a part of? Obviously I don't want to be losing the features I've gained when I upgraded from 2.4 to 2.5 and I am looking forward to some of the newer 2.6 features as well.

PR #662, opened 2016-09-08
Fix serving static resources in WildFly

mthmulders:
While building a Spark app that will be deployed in WildFly 10, I noticed that static resources didn't work. It turns out that WildFly doesn't give a `jar://` URL when loading a resource from the classpath, but instead the URL starts with `vfs://`. Loading the resource using `getResourceAsStream` later on works as expected, so the only thing that needed changing was a simple check on the resource URL protocol.


mthmulders:
Sure thing, and I'd definitely do that... if not because `src/test/java/spark/staticfiles/StaticFilesTest.java` is completely commented... A quick scan revealed that there's more in that class than just removing the comment signs.

tipsy:
This has been refactored a lot, so closing. Feel free to reopen if it's still a problem.

PR #665, opened 2016-09-16
add hasKey method

chrisgreve:
I would like to add a hasKey (or containsKey if you like) method to QueryParamsMap. The use case is for a Boolean queryParam entry that has a default value. An example would be:

url/?describe
or
url/?describe=true

being equivalent.  Adding the method makes for cleaner code and removes the construction of a new map to call queryParamsMap.toMap().containsKey()


perwendel:
tx

Issue #619, opened 2016-07-12
content-type of static files is always "application/octet-stream"


martin-pabst:
Static files are always served with content-type "application/octet-stream". Content type should be inferred from filetype.

tipsy:
Fixed!

PR #667, opened 2016-09-22
Refactoring of JarResourceHandler

totomz:
In this pr I refactored the `JarResourceHandler` class to handle static files from within a Jar archive ("jar resources"), in the same way of ExternalFileResource. 

The previous implementation did not extend `AbstractResourceHandler`, so in     `StaticFilesConfiguration.java` there were 2 different methods for handling static files:

``` java
private boolean consumeWithFileResourceHandlers(HttpServletRequest httpRequest,
                                                    HttpServletResponse httpResponse) throws IOException {
        if (staticResourceHandlers != null) {

            for (AbstractResourceHandler staticResourceHandler : staticResourceHandlers) {

                    httpResponse.setHeader(MimeType.CONTENT_TYPE, MimeType.fromResource(resource));
                    customHeaders.forEach(httpResponse::setHeader); //add all user-defined headers to response
                    OutputStream wrappedOutputStream = GzipUtils.checkAndWrap(httpRequest, httpResponse, false);
                    IOUtils.copy(resource.getInputStream(), wrappedOutputStream);
                    wrappedOutputStream.flush();
                    wrappedOutputStream.close();
                    return true;
                }
            }

        }
        return false;
    }
```

and 

``` java
    private boolean consumeWithJarResourceHandler(HttpServletRequest httpRequest,
                                                  HttpServletResponse httpResponse) throws IOException {
        if (jarResourceHandlers != null) {

            for (JarResourceHandler jarResourceHandler : jarResourceHandlers) {
                InputStream stream = jarResourceHandler.getResource(httpRequest);

                if (stream != null) {
                    // <-- MISSING CONTENT-TYPE HERE -->
                    OutputStream wrappedOutputStream = GzipUtils.checkAndWrap(httpRequest, httpResponse, false);
                    customHeaders.forEach(httpResponse::setHeader); //add all user-defined headers to response

                    IOUtils.copy(stream, wrappedOutputStream);

                    wrappedOutputStream.flush();
                    wrappedOutputStream.close();

                    return true;
                }
            }
        }
        return false;
    }
```

The duplication introduced a bug in serving static files from the JAR archive, since no Content-Type was set for them (#618 and #619 did not fix the issue for JAR resources, only for external files)

In this pull request, the method `consumeWithJarResourceHandler` has been removed, and the Jar resources are threated as ResourceFileHandler. A specific test for this functionality is missing, because to test sould run against a compiled archive, requiring major modification to the pom and to arquillian. Of course, I can do that (but I would like to discuss the hows and whats :)


perwendel:
JarResourceHandler has been removed.
Thanks for contributing!

PR #671, opened 2016-09-30
SPARK-193 Adds a servlet-based routing mechanism to complement the existing filter-based one.

### LINKED ISSUE #193: Why did you use a Filter instead of a Servlet?

coyotesqrl:
Fix #193 

Instead of deprecating the current filter-based mechanism in favor of a servlet-based one (as in [this PR](https://github.com/perwendel/spark/pull/254)), this keeps both and allows users to choose based on their needs.

Our environment needs servlet-based routing; until it is directly incorporated into Spark we'll need to maintain our own `SparkServlet`.


coyotesqrl:
It had been quite awhile since this PR had been created; I thought it was about time to rebase with master.

coyotesqrl:
Rebasing again, as this PR has now been open 10 months and I want to avoid any potential drift or conflicts.

src/test/java/spark/servlet/WrapperRequestTest.java
```diff
@@ -0,0 +1,129 @@
...
+        executorService.submit(() -> {
+            try {
+                LOGGER.info(">>> STARTING EMBEDDED JETTY SERVER for jUnit testing of SparkFilter");
+                server.start();
+                latch.countDown();
+                System.in.read();
+            } catch (Exception e) {
+                e.printStackTrace();
+                System.exit(100);
+            }
```
jakaarl: Why on Earth a call to `System.exit()`?

src/test/java/spark/servlet/WrapperRequestTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+        executorService = Executors.newSingleThreadExecutor();
+        executorService.submit(() -> {
+            try {
+                LOGGER.info(">>> STARTING EMBEDDED JETTY SERVER for jUnit testing of SparkFilter");
+                server.start();
+                latch.countDown();
+                System.in.read();
+            } catch (Exception e) {
+                e.printStackTrace();
```
jakaarl: Using `Exception.printStackTrace()` is _very_ seldom justified. Why not propagate exception instead?

src/main/java/spark/servlet/SparkServlet.java
```diff
@@ -0,0 +1,273 @@
...
+            } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+                LOGGER.debug(
+                        "Error converting {} to BiFunction<HttpServletRequest, String, HttpServletRequestWrapper>; "
+                                + "falling back to default",
+                        wrapperSupplierName,
+                        e);
+            }
+
+        }
+
```
jakaarl: Is it justified to fall back to default instead of crashing and burning?

src/main/java/spark/servlet/SparkServlet.java
```diff
@@ -0,0 +1,273 @@
...
+    private SparkApplication getApplication(String applicationClassName) {
+        try {
+            Class<?> appClass = Class.forName(applicationClassName);
+            if (SparkApplication.class.isAssignableFrom(appClass)) {
+                return SparkApplication.class.cast(appClass.newInstance());
+            }
+        } catch (ClassNotFoundException | InstantiationException | IllegalAccessException e) {
+            LOGGER.debug("Error converting {} to SparkApplication", applicationClassName, e);
+        }
+
```
jakaarl: This is way lenient - personally I'd prefer both the `isAssignableFrom()` check and caught exceptions to raise hell to just going on and merely logging at debug level. Being more strict would also get rid of the potential null return value and `Objects.nonNull()` that handles it.

src/main/java/spark/servlet/SparkServlet.java
```diff
@@ -0,0 +1,273 @@
...
+        try {
+            path = URLDecoder.decode(path, "UTF-8");
+        } catch (UnsupportedEncodingException ignore) {
+            // this can't really ever happen
+        }
+
+        LOGGER.debug("Relative path = {}", path);
+        return path;
+    }
+
```
jakaarl: Do the above two methods really need to be static?

jakaarl:
I think `Servlet` would be preferable to `Filter`, and would like to see this PR merged. I do have some reservations about the implementation (wrote some comments to the changes made), but they're mostly matters of opinion and style.

src/main/java/spark/servlet/SparkServlet.java
coyotesqrl: That's a preference, that methods that do not read/write state be static.

src/main/java/spark/servlet/SparkServlet.java
coyotesqrl: My feeling is that it is safe to do so, as missing configuration is not something that would randomly occur at runtime but be consistently missing/wrong during the development cycle. However, I could as easily make the inverse argument, that since it can be resolved by fixing the servlet configuration early, a hard fail here would force that to be done. I can go either way.

src/main/java/spark/servlet/SparkServlet.java
coyotesqrl: I can see that point of view. I lean toward a more lenient environment - as should be clear from the way I'm falling back to the default wrapper supplier above - with logging; however, I'm not opposed to a stricter mode.

src/test/java/spark/servlet/WrapperRequestTest.java
coyotesqrl: Because it's only a unit test, I took a more aggressive approach of cleaning up and logging in the case where the embedded Jetty server does not come up correctly. To do it the right way would look more like...
```java
try {
    LOGGER.info(">>> STARTING EMBEDDED JETTY SERVER for jUnit testing of SparkFilter");
    server.start();
    latch.countDown();
    System.in.read();
} catch (Exception e) {
    try {
        latch.countDown();
        server.stop();
    } catch (Exception e1) {
        throw new RuntimeException(e1);
    }
    throw new RuntimeException(e);
}
```

I just didn't think the extra boilerplate was worthwhile for it only being a test.

coyotesqrl:
Rebased and pushed. I am still hopeful that this PR - or a similar solution - can be prioritized for #193 and eventually merged into the product. I would very much like to stop maintaining my own version of `SparkServlet`.

coyotesqrl:
Rebased with master and made a few small concurrency updates to the `SparkServlet`.

coyotesqrl:
Rebased again and resolved merge conflict.

As this PR has been open for over two years, it would be great to have an ETA on when a decision on whether to accept or reject the change will be made.

tipsy:
@coyotesqrl, I'm not supposed to speak on behalf of the project, but since you've been putting so much effort into this I'll make an exception. I think it's unlikely that this PR will be merged.

As you can see from the git history there hasn't been a lot of activity lately. We've been discussion Spark 3 a bit, but not much progress has been made. At any rate, adding major changes to Spark 2.X seems unlikely at this point. I hope we can use the work I've been doing on [Javalin](https://javalin.io/) as a base and build Spark 3 on top of it, but nothing has been decided.

PR #672, opened 2016-10-01
Uri Not Found Handler

fcambarieri:
Hi,

```
 I'm adding a way to handle pages not found. The configuration is similar at the way that it's mapped the exceptions.

 For Rest API could be very useful.
```

Best Regards.


tipsy:
Thanks! Similar functionality was introduced in #719

PR #683, opened 2016-10-16
test content type for static files from an archive

reinhard:
Add a test case that checks the content type of a resource that is being read from a jar. Since this test requires an additional file on the classpath, I have to set up a special classloader that adds this file. Then this classloader can be used to setup Spark.


tipsy:
Thanks!

Issue #688, opened 2016-10-21
Priority of accept types is incorrect if the client specifies no q factor


staticfinal:
I have multiple routes for the same URL with different accept types:

```
        Spark.get("/channel/:version/", new GetChannelFeed());
        Spark.get("/channel/:version/", "text/html", new GetChannelFeedHtml(), templateEngine);
```

I get the HTML version of the channel in all browsers except IE. There I get the RSS feed. The reason is that IE sends an Accept header without q factor < 1 for "*/*".

The full accept header for IE is "text/html, application/xhtml+xml, */*". So both the text/html Route and the */* Route match with the same quality.

But the standard is pretty clear that in this case the most specific media range should be preferred:

> If more than one media range applies to a given type, the most specific reference has precedence.
> https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

So I should get the HTML Version and not the RSS feed.

Chauncey-Xxy:
I will try to fix it.

PR #689, opened 2016-10-21
Try fix for bug 688

staticfinal:
This is a very simple and hacky fix for bug 688.


tipsy:
Thanks @staticfinal, but I think this can't be merged, as people might be relying on the other (incorrect?) behavior. We'll keep the #688 open and fix it in 3.0.

PR #692, opened 2016-10-28
Make RouteImpl factory methods public

kliakos:
Same as factory methods of `ResponseTransformerRouteImpl`.

Otherwize you can't use the `Service#addRoute()` method


kliakos:
Can we merge this one please?

tipsy:
Sure, why not.

Issue #700, opened 2016-11-04
Arbitrary File Read Vulnerability


RichoDemus:
Just posting it here so it gets visbility, I didn't write the original message:  
http://marc.info/?l=full-disclosure&m=147814643630342&w=2

There appears to be a vulnerability which lets users read any file from the file system

jakaarl:
Eeek! 8-O
If the core folks aren't available, I could take a stab at this tonight/tomorrow.

nightwatchcyber:
CVE-2016-9177 has been assigned for this vulnerability:

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9177

jakaarl:
I wasn't able to reproduce in a minimal Spark app JAR, but it's indeed reproducible by running `spark.examples.staticresources.StaticResources` and curling away at it. Will have to investigate more.

jakaarl:
Not really having the time and peace to thoroughly investigate, but based on some quick tests:
 * from IDE, both class path and external file resources are vulnerable
 * running in a standalone Jetty JAR, only external resources are accessible

The difference in running in an IDE and running as a packaged up is probably either due to resources being inside a JAR file, or different class loader setup. Should try running in an unpackaged application.

perwendel:
We are available. The emails "AJ" tried to send all got caught by gmail's spam filters.
We are addressing this ASAP.

perwendel:
Fixed with #702.
Spark 2.5.2 released http://search.maven.org/#artifactdetails%7Ccom.sparkjava%7Cspark-core%7C2.5.2%7Cbundle with this fix!

mattwelke:
Noticed my IDE warned me about the vulnerability after I followed the tutorial on the site. The tutorial uses version `2.5` right now.

PR #701, opened 2016-11-05
Fix for #700 - Arbitrary File Read Vulnerability

perwendel:
Fix for #700 
This is my take on the directory traversal problem.
Please review and try out.


perwendel:
I've just verified it for Windows, didn't have time to test for linux etc

jakaarl:
DISCLAIMER: Didn't really have time to look closely, so I could be talking out of my arse so to speak, but: wouldn't using `File.getCanonicalPath()` be safer than string manipulation? I'd certainly trust JDK library methods more than my own code, especially when it comes to security. Also, there are all kinds of clever ways (encodings, escapes and what not) of defeating checking strings for "..", "/", "\" etc.

perwendel:
@jakaarl Sorry, merged before I saw your comment. The guy making the post emailed and said it was a good solution. But I'll follow your advise and see if I can make it better. New PR will be created if you're right!

PR #704, opened 2016-11-08
Workaround for issue 684

paul-hammant:
Tests demonstrate what you have to do with 2.5.1 and above to use suffixless static files, to avoid default mime type of 'application/octet-stream'


tipsy:
Thanks, but I don't think there's much point in merging this now(?)

Issue #703, opened 2016-11-07
Error with static files in version 2.5.2


giflw:
Hi

I was using v2.5.0, but static files content-type header was missing (v2.5.1 has the same issue).

Then I switched to 2.5.2, but my static files wasn't loading, and I get traversal directory warning from StaticFilesConfiguration.

tipsy:
Unix?

perwendel:
That sucks! What OS are you using? I verified functionality in Windows and Linux. All information you provide is helpful!

giflw:
Windows 8, Java 8 Oracle

perwendel:
Is it possible to describe your use? Is it internal or external static file location? What does the path look like? What exactly is not working?

giflw:
I'm using external files, with a folder named 'tools' and registered paths '/tools', '/tools/:id'. There are resources, like images, acessible from '/tools/1/image.pg'. The imagem is supposed to load, as there are no route matching the pattern, but I receive the traversal director warning and the imagem doesn't load.

giflw:
I had a /* route too

perwendel:
ok, thanks, I'll try to reproduce

perwendel:
@giflw 
I did the following test:

```
import static spark.Spark.get;
import static spark.Spark.staticFiles;

public class StaticResources {

    public static void main(String[] args) {

        String directoryRoot = System.getProperty("java.io.tmpdir") + "external\\";
        System.out.println(directoryRoot);

        staticFiles.externalLocation(directoryRoot);

        get("/tools", (request, response) -> "Hello Tools!");
        get("/tools/:id", (request, response) -> "Hello Tools: " + request.params(":id") + "!");
        get("/*", (request, response) -> "Hello Everyone else!");
    }
}
```
directoryRoot = C:\Users\<myusername>\AppData\Local\Temp\external\

Where I created a subfolder in the folder with path directory root name `tools` and a folder in that folder named `1` and a file in that folder name `image.png`.

`http://localhost:4567/tools/1/image.png` - returned the image
`http://localhost:4567/tools` - returned "Hello Tools!"
`http://localhost:4567/tools/hammer` - returned "Hello Tools: hammer!"
`http://localhost:4567/tools/hammer/ish - returned "Hello Everyone else!"

Is this test analog to what didn't work for you? If so can you try this out and see if it fails?

giflw:
This is analog to what I did, but the image wasn't returned. I'll try this code as soon as I can.

[...]

PR #706, opened 2016-11-08
Make external static folder be *nix like

giflw:
When Windows paths are received, like `C:\Users\USERNAME\AppData\Local\Temp\external\`, they will be changed to `C:/Users/USERNAME/AppData/Local/Temp/external`.

### DIFF
--- a/src/main/java/spark/staticfiles/DirectoryTraversal.java
+++ b/src/main/java/spark/staticfiles/DirectoryTraversal.java
@@ -1,6 +1,7 @@
 package spark.staticfiles;
 
 import static spark.utils.StringUtils.removeLeadingAndTrailingSlashesFrom;
+import java.nio.file.Paths;
 
 /**
  * Protecting against Directory traversal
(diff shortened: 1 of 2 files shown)


giflw:
Tested with code bellow as in #703 :

```java
import static spark.Spark.get;
import static spark.Spark.staticFiles;

public class StaticResources {

        //String directoryRoot = "/Users/USERNAME/AppData/Local/Temp/external/";
        //String directoryRoot = "c:/Users/USERNAME/AppData/Local/Temp/external/";
        String directoryRoot = System.getProperty("java.io.tmpdir") + "external\\";

        staticFiles.externalLocation(directoryRoot);

        get("/tools", (request, response) -> "Hello Tools!");
        get("/*", (request, response) -> "Hello Everyone else!");
    }
}

perwendel:
Starting review!

perwendel:
this looks good! I'm going to test it out with before merging!

giflw:
Nice! Thanks =D

perwendel:
Thanks for fixing man!

giflw:
HI
I know that was already merged, but I was thinking about the **performance implications** of `String nixLikePath = Paths.get(path).toAbsolutePath().toString().replace("\\", "/");` in all file requests. 

Maybe just setting the external location path in the O.S. dependet format would suffice. So, in the registration the path would be transformed by  `Paths.get(folder).toAbsolutePath().toString()` and in the directory traversal check would be just the `path`, as it was taken from a `File` object in `ExternalResource` constructor  (`new File(StringUtils.cleanPath(path));`) and the path is returned using `file.getPath` (which is O.S. dependent).

If you wanna try this approach, I can update the PR.

Thanks

PR #707, opened 2016-11-09
Add path-prefix functionality

tipsy:


tipsy:
Usage:
```
path("/api", () -> {
    post("/add-role",        Api.addRole);
    post("/add-email",       Api.addEmail);
    delete("/remove-email",  Api.addMsisdn);
});
```

tipsy:
Any thoughts on this @perwendel? @jakaarl? @OpsRJ?

jakaarl:
I love the usage, very fluent.

src/main/java/spark/Service.java
```diff
@@ -71,6 +73,7 @@
     protected Optional<Integer> webSocketIdleTimeoutMillis = Optional.empty();
 
     protected EmbeddedServer server;
+    protected Stack<String> pathStack = new Stack<>();
```
jakaarl: I think `Stack` is one of those "not recommended" classes, quoting:
> A more complete and consistent set of LIFO stack operations is provided by the Deque interface and its implementations, which should be used in preference to this class. 

Stack is, IIRC, synchronized and eg. `ArrayDeque` probably not, but from what I understood (took a quick glance only, though) there is no need for it here. And if there was, using `Stack` wouldn't really close the window of opportunity for a race condition.

tipsy:
Your comment vanished when I amended the commit, sorry about that. I have change to Deque now, as well as added some more tests (filters, triply nested paths) and documentation.

MouettE-SC:
👍 like it

OpsRJ:
@tipsy Very nice. Loved it. Thanks!

perwendel:
Good stuff!

PR #708, opened 2016-11-13
Guava's abstract service for awaiting stopped

kdubezerra:
This patch proposes to implement Spark's Service class
as an AbstractService (from Guava), which provides a
simple API to have both synchronous and asynchronous
control of the service life cycle (start, stop, etc.).

Ref: github.com/perwendel/spark/issues/705

Signed-off-by: Eduardo Bezerra <kdubezerra@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,12 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>20.0</version>
+        </dependency>
+
         <!-- LOGGING DEPENDENCIES -->
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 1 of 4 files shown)


tipsy:
I don't think @perwendel will accept Guava as a dependency.

kdubezerra:
@tipsy Oh, if that's the case, I won't bother fixing the PR... I think @jakaarl is already working on a blocking stop() method, and for my use case, that would suffice. I just thought it might be useful to follow guava service's api.

jakaarl:
@kdubezerra I'll try get stuff done in the next couple of days - kind of dropped working on it. I have the simple solution pretty much done, I guess it'll suffice for now and I'll return to the problem when I have more time and energy to spend. Will check out the Guava stuff for inspiration.

PR #719, opened 2016-12-01
Custom 404 and 500 error pages - Proposal

perwendel:


src/main/java/spark/CustomErrorPages.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public static Object getFor(int status, Request request, Response response) {
+
+        Object customRenderer = CustomErrorPages.getInstance().customPages.get(status);
+        Object customPage;
+
+        if (customRenderer instanceof String) {
+            customPage = customRenderer;
+        } else {
```
tipsy: Change to `if(customRenderer instanceof Route)` ?

src/main/java/spark/CustomErrorPages.java
```diff
@@ -0,0 +1,75 @@
...
+            try {
+                customPage = ((Route) customRenderer).handle(request, response);
+            } catch (Exception e) {
+                customPage = status == 400 ? NOT_FOUND : INTERNAL_ERROR;
+            }
+        }
+
```
tipsy: Remove line?

src/main/java/spark/RequestResponseFactory.java
```diff
@@ -26,6 +26,13 @@
     private RequestResponseFactory() {
     }
 
+    /**
+     * Solely used for custom error pages.
```
tipsy: I'm not sure this comment is helpful. Could the method be used for anything else? If not, call it something that contains the words "custom error pages", and remove the comment?

src/main/java/spark/Service.java
```diff
@@ -323,6 +323,39 @@ public synchronized Service webSocketIdleTimeoutMillis(int timeoutMillis) {
...
+    }
+
+    /**
+     * Maps 500 internal server errors to the provided route.
+     */
+    public void internalServerError(Route route) {
+        CustomErrorPages.add(500, route);
+    }
+
+
```
tipsy: 2x empty  line

src/main/java/spark/RequestResponseFactory.java
perwendel: It's for specifying that it only should be used from there. But when thinking about it this class should no nothing on which classes that uses it. I'll re-write it

src/main/java/spark/CustomErrorPages.java
perwendel: done

src/main/java/spark/CustomErrorPages.java
perwendel: done

src/main/java/spark/RequestResponseFactory.java
perwendel: fixed

src/main/java/spark/Service.java
perwendel: fixed

PR #726, opened 2016-12-05
Add pluggable RequestLog to embedded Jetty server for use with loggers such as slf4j, log4j, etc.

ygaller:
Add option to specify a request log for embedded Jetty to use.

Format:
requestLog(RequestLog.class); 

this function should be called before any route mappings (similar to the websocket(...) call).
 
The provided logger should implement org.eclipse.jetty.server.RequestLog. This will allow logging of all requests, via the console, slf4j or log4j. For example, via loggers such as NCSARequestLog, Slf4jRequestLog, etc.

Sample output when running the RequestLogExample test class, using the Slf4jRequestLog:

[qtp29583191-17] INFO org.eclipse.jetty.server.RequestLog - 0:0:0:0:0:0:0:1 - - [05/Dec/2016:20:30:53 +0000] "GET //localhost:4567/log HTTP/1.1" 200 7

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -29,6 +29,9 @@ import org.slf4j.LoggerFactory;
 
 import spark.embeddedserver.EmbeddedServer;
 import spark.embeddedserver.EmbeddedServers;
+import spark.embeddedserver.jetty.logger.RequestLogClassWrapper;
+import spark.embeddedserver.jetty.logger.RequestLogInstanceWrapper;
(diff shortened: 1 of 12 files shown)


jakaarl:
I think this is a welcome feature. Haven't looked properly into the changes, but at least the asterisk imports should be fixed?

ygaller:
@jakaarl Rolled back the asterisk import changes (IntelliJ import optimization...).

s4mpl3d:
Feature would be very helpfull && welcome

tipsy:
Should be solved by #813

Issue #705, opened 2016-11-08
Need a Spark.awaitStopped() method


kdubezerra:
The `Spark.stop()` method stops the server, setting the internal `initialized` variable to `false`. The problem is that this is all done asynchronously, in a separate thread (in the `Spark.Service` class):

```
/**
 * Stops the Spark server and clears all routes
 */
public synchronized void stop() {
    new Thread(() -> {
        if (server != null) {
            routes.clear();
            server.extinguish();
            latch = new CountDownLatch(1);
        }

        staticFilesConfiguration.clear();
        initialized = false;
    }).start();
}
```

However, I found no way of checking whether the service has actually stopped. This is problematic because Spark creates a singleton service, so shutting down a server and restarting it without actually finishing the server process may lead to problems. In a more concrete use case, this creates problems when running several unit tests, where each unit test starts a server, runs the test, and shuts the server down.

This issue would be solved by providing a method like `Spark.awaitStopped()` that returned only when the thread started in `Spark.Service.stop()` had finished.

jakaarl:
I did some quick experimenting, shouldn't be too hard to implement.

jakaarl:
My quick hack of an implementation basically does a `join()` on the thread if waiting is required. Sound okay (@perwendel @tipsy)? Need to write tests at least before raising a PR.

kdubezerra:
I think a better way to go would be with a condition variable with a lock on the `initialized` field... At the end of the `stop` thread, you'd signal the condvar. The awaitStopped() would get the lock, check the value of `initialized`, waiting on the condvar if necessary, then releasing the lock.

jakaarl:
@kdubezerra sounds more complicated than necessary, what benefits would the added complexity bring?

kdubezerra:
@jakaarl That's a good question. To me, it looks like a textbook example of when a condition variable should be used: all it depends on is the `initialized` variable being false (hence the condition). And, fwiw, using the condvar will consist of ~5 (10 at most) lines of code. But maybe using a `Thread.join` would be just as correct and probably simpler.

I have another question though: what if you call `Spark.awaitStopped()` before starting the server? What if it's called after the Thread spawned by `Spark.stop()` has already finished? Do you keep a reference to that thread just in case someone calls `awaitStopped()`?. My 2 cents is that the condvar would cover all these cases, but I'm just as happy with whatever implementation you guys provide :)

jakaarl:
[...]

Issue #563, opened 2016-05-19
Calling stop() causes java.lang.InterruptedException


reynders:
On OSX El Capitan using SparkJava 2.5 expanding on the sparkjava.com hello world example:

    import static spark.Spark.*;

    public class SparkTest {
        public static void main(String[] args) {
            get("/hello", (req, res) -> "Hello World");

            get("/stop", (req, res) -> {
                stop(); 
                return "doh";
            }); 
        }
    }

using browser to hit first hello -> Hello World, then stop -> 15 seconds of silence and then the server throws this up:

20:27:26.448 [qtp276637015-15] ERROR spark.embeddedserver.jetty.EmbeddedJettyServer - stop failed
java.lang.InterruptedException
	at java.lang.Object.wait(Native Method) ~[?:1.8.0_65]
	at java.lang.Thread.join(Thread.java:1253) ~[?:1.8.0_65]
	at org.eclipse.jetty.util.thread.QueuedThreadPool.doStop(QueuedThreadPool.java:161) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.jetty.util.component.AbstractLifeCycle.stop(AbstractLifeCycle.java:89) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.jetty.util.component.ContainerLifeCycle.stop(ContainerLifeCycle.java:143) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.jetty.util.component.ContainerLifeCycle.doStop(ContainerLifeCycle.java:161) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.jetty.server.handler.AbstractHandler.doStop(AbstractHandler.java:73) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.jetty.server.Server.doStop(Server.java:476) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
	at org.eclipse.jetty.util.component.AbstractLifeCycle.stop(AbstractLifeCycle.java:89) ~[spark-test-1.0-SNAPSHOT-jar-with-dependencies.jar:?]
[...]

PR #730, opened 2016-12-06
Added Service.stopAndWait().

jakaarl:
A simple blocking `stop()` method. See #705 .


jakaarl:
I'll add a test. From the top of my hat, I think there shouldn't be a problem - it still spawns a new thread to handle stopping, but in the sync case, blocks until completion.

jakaarl:
If the problem really is the fact that a request handler thread waits on the thread stopping the server, which waits on its handler threads to finish, this could be damn hard to fix (especially without taking a stab at the server sources, but that's not really an option). @perwendel any suggestions how to tackle this? Just have a huge big fat warning in the doc comments about calling the blocking stop from handlers? Give up? Other, please specify: ____________ ?

fdmoulin:
Personally, I don't think that the issue raised in https://github.com/perwendel/spark/issues/563 is something that was broken to begin with.  Invoking stop directly via the server's own request handler should inevitably result in the server not being able to complete the response if you really think about it.  IMHO it would have been acceptable to document this as a non-recommended use.  The fact that a new thread was added to deal with it makes the framework more forgiving in case someone does this, but it still doesn't make it a good design choice to invoke stop directly from within a request handler.

In the case where someone wants to stop their application with its own API service (which happens to be a use case in my own project btw) it would be better to have the API service set a shutdown flag, and have the main application thread perform a graceful shutdown once the server response has completed.

The downside of the 563 fix is that the stop method is always non-blocking.  This can cause issues for the "usual" use cases where stop isn't invoked form the request handler.  And this is why someone asked to have a blocking stop re-added.

I would personally consider it acceptable to have both methods (a blocking and non-blocking), and **_document them really well_**.  So add your proposed stopAndWait() method, and leave the existing stop as-is.

perwendel:
Starting reviewing this PR this week!

jakaarl:
I think implementing a dedicated method for "stop request" would be worth considering. Maybe just a wrapper around Jetty's shutdown handler: http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/server/handler/ShutdownHandler.html

Clearly this would warrant an issue and PR of its own.

bwssytems:
It looks like this is a good solution, but there is still an issue if you want to re-init after a stop.

If there is a staticFileLocation("path") call, this gets cleared on the stop but cannot be reset do to this code:

```
    public synchronized Service staticFileLocation(String folder) {
        if (initialized && !isRunningFromServlet()) {
            throwBeforeRouteMappingException();
        }

        staticFileFolder = folder;

        if (!servletStaticLocationSet) { /*** this needs to be fixed when a stop happens ***/
            staticFilesConfiguration.configure(staticFileFolder);
            servletStaticLocationSet = true;
        } else {
            LOG.warn("Static file location has already been set");
        }
        return this;
    }

```

eldargab:
So, when this will be merged?

sharebear:
Just adding another use-case for this. In order to perform a graceful shutdown when running as a docker container the process is supposed to handle the SIGTERM signal. I can add a signal handler that calls Spark.stop() but due to it's asynchronous nature the JVM stops before Jetty has a chance to gracefully shutdown in-flight requests.

This causes a problem when rolling out new versions of my software as all in-flight requests (we have quite a lot at any given time) get dropped on the floor rather than running to completion.

kuenishi:
I'd like to give my +1 to this PR. As @fdmoulin mentioned "document them really well" , behavior of stopAndWait could be carefully documented in sparkjava.com , or could be left undocumented? I'm OK with it undocumented because I just use synchronous stop for unit testing.

jakaarl:
@bwssytems so the problem is, that `servletStaticLocationSet` is not cleared on stop? It's kind of redundant, since the flag is duplicated in `StaticFilesConfiguration` - perhaps I should refactor so that the configuration object is the single source of truth...

jakaarl:
@bwssytems @tipsy @perwendel could you take a look at PR #902 ? It would be needed to address the issue with stop and clearing static file configuration.

bwssytems:
@jakaarl That resolution should be good. The issue was that on a restart, the static directory location is already there but it gets cleared on the stop and then cannot be reset and takes the error of no folder location.

gdunkle:
+1 on this PR it would be really helpful

Tomtomgo:
Ran into the same issue: https://stackoverflow.com/questions/46451309/how-to-wait-until-spark-service-is-stopped.

So definitely a +1 on the requested functionality.

I do think that given we initialize like this:

```java
Spark.init();
Spark.awaitInitialization();
```

It makes more sense to implement:

```java
Spark.stop();
Spark.awaitStop();
```

qmg-aki:
@perwendel any updates on this? It would be great to have that functionality in.

jakaarl:
Refactored the PR to resemble `init()` and `awaitInitialization()` as suggested.

I have a feeling the threading stuff in both init and stop could be reworked to be clearer and possibly safer - currently it makes me slightly uneasy.

simonreye:
So has/will this be merged and part of a release?

perwendel:
Good stuff! Sorry for the lag time!

simonreye:
Can we expect an official release soon with this change in it?

MrThreepwood:
Same question. I guess I'm not sure if this would fix my problem, but at I'm trying to figure out how to gracefully shutdown my server (IE: Let all current requests complete, then shutdown, rather than killing  everything in process.)

tipsy:
This method doesn't really help with that. You need to give Spark a custom Jetty server which is configured for graceful shutdown.

jakaarl:
IMHO, graceful shutdown should be the default, or at least easily configurable. Agree?

There's some useful tips here: https://github.com/eclipse/jetty.project/issues/2076

tipsy:
Yes, I agree.

ggrnd0:
Is it already in 2.8.0?

PR #732, opened 2016-12-07
Fix line endings on source files

MouettE-SC:
Some files have been commited in repository with CRLF line endings whereas the vast majority of files have LF line endings.


tipsy:
Thanks!

MouettE-SC:
Thanks that's great, it will make custom branching and merging much easier.

Issue #568, opened 2016-05-26
Support multiple calls to staticFileLocation


jylin:
It seems you currently can't call staticFileLocation twice (I didn't look at the source code, but it looks like the second call is ignored?).

I'd like to be able to have a static assets folder, but also use webjars. However, something like this doesn't seem to work:
        staticFileLocation("/public");
        staticFileLocation("/META-INF/resources");

It doesn't look like I can't access the files in "/META-INF/resources", but if I comment out the first line, then those files are successfully loaded.

vacax:
this issue is important.

pgrisafi:
+1
New to Spark. As a webjars fan, this is important.
A quick and dirty hack I made, using a custom route and a guava cache
Disclaimer: It works for me, maybe not for you. It does not handle multiple nested folders, but it is easy to extend to do so.

	private static final LoadingCache<String, String> WEBJARS_CACHE = CacheBuilder.newBuilder()
		       .maximumSize(1000)
		       .expireAfterWrite(365, TimeUnit.DAYS)
		       .build(
		           new CacheLoader<String, String>() {
		        	   @Override
		             public String load(String fullPath) throws IOException {
		            	 InputStream inputStream = Application.class.getClassLoader().getResourceAsStream(fullPath);
		     			return IOUtils.toString(inputStream);
		             }
		           });
	
	public static void main(String[] args) {
		staticFiles.location("public");
		get("/hello", (req, res) -> "Hello World");
		
		get("/webjars/:library/:version/:file",  (req, res) -> {
			String library = req.params("library");
			String version = req.params("version");
			String file = req.params("file");
			if(file.endsWith(".js")){
				res.type("application/javascript");   
			}
			if(file.endsWith(".css")){
				res.type("text/css");   
			}
			String fullPath = "META-INF/resources/webjars/" + library + "/" + version +"/" + file;
			return WEBJARS_CACHE.get(fullPath);
		});
	}
I can access files using links like

    <script src="/webjars/angularjs/1.5.6/angular.min.js"></script>
    <link rel="stylesheet" href="/webjars/angular-material/1.1.0-rc4/angular-material.min.css">

AzureusNation:
I'm looking to implement this for you but could you give me a bit more detail in what the goal is?

drxaos:
I have multiple jars with resource folder and I had to do some reflection magic to search static resources in all jars.
[...]

PR #734, opened 2016-12-13
fix: Support multiple calls to staticFileLocation #568

udonne:
You can add multiple static file locations like:
staticFiles.location("/public");
staticFiles.location("/secondPublic");


atomfrede:
Having this would be really great as you  could use custom files from `static` and webjars together.

tipsy:
@udonne This looks great, could you update it?

tipsy:
Pinging @udonne

BloodShura:
Any updates?

Issue #543, opened 2016-05-06
Improve multipart handling (file-uploads)


tipsy:
Currently you have to do:
```java
request.attribute("org.eclipse.jetty.multipartConfig", new MultipartConfigElement("/temp"));
request.raw().getPart("uploaded_file") // "uploaded_file" is the html input name
```

While this isn't terrible, it could be prettier. Something like:
```java
request.enableMultipart() // could call request.enableMultipart("/temp")
request.part("uploaded_file")
```

kshep92:
Has anyone submitted a pull request for this yet?

tipsy:
Nope. I have one locally, where I just have a .part() method that upgrades the request automatically, but it only works with Jetty. I'm not sure how uploads work with other webservers, which is why I'm holding off on publishing it.

kshep92:
I guess until you're done, we could use a filter to do this for us.

PR #754, opened 2017-01-30
New abstraction for enabling/using multipart

rob-signorelli:
This is a contribution for issue #543. While the issue asked for a more seamless way to enable multipart, I took it a step further and encapsulated the whole multipart experience into a new abstraction. Instead of interacting w/ the raw Jetty/servlet "Part" instance, I've added a separate class that lets users stay at the Spark level.

```java
Spark.post("/my/file/upload", (req, res) => {
    // Not required, but part() returns null and parts() returns empty if not true
    if (!req.isMultpart()) {
        throw new RuntimeException("Not a multipart content type");
    }

    try (MultipartPart foo = req.part("foo.txt");
         MultipartPart bar = req.part("bar.png")) {
        String fooText = foo.asText();
        InputStream barStream = bar.inputStream();  // auto-closed when part is closed via try w/ resources
       ... do something interesting ...
    }
});
```

There are 3 ways to access parts

```java
// Grab the part/file w/ the specific part name
MultipartPart foo = req.part("foo");

// Multiple parts/files uploaded w/ the same part name
Collection<MultipartPart> allBars = req.parts("bar");

// Grab all parts/files from the request
Collection<MultipartPart> everything = req.parts();
```

Some of the benefits of this approach/implementation:

* The jetty request attribute is automatically lazy-set when you make your first call to ```part()``` or either of the ```parts(String)``` overloads. One less thing for the programmer to think about. You can specify your own value if you care to beforehand but it's no longer a requirement as it will default to "/temp" just like in the docs.
* MultipartPart is auto closable so we can use try-with-resources to better make sure that we're cleaning up all of our data streams when the request is finished.
* Removes the need to dip into the raw servlet request. Perhaps I'm a purist but the more I can stay in Spark-land and not have to dive into Jetty/Servlet-land the better off I feel. The additional request methods match Spark's existing style for interacting with the request so it feels more natural to me.
* Not sure if it's a desire for Spark to support pluggable HTTP engines at some point in the future, but if so, this implementation is no longer as tightly coupled to Jetty's implementation. Between the ```MultipartConfig``` class and ```MultipartPart``` being an interface w/ a servlet reference implementation, it should be easy to swap out web server specific implementations should you choose to do that.

I've included a series of unit tests for these classes and updated the tests for the util classes I updated. Hopefully you find this a useful addition to Spark.

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ public class JsonAcceptTypeExample {
 ```
 ---------------------------------
 
-Example showing how to render a view from a template. Note that we are using `ModelAndView` class for setting the object and name/location of template. 
+Example showing how to render a view from a template. Note that we are using `ModelAndView` class for setting the object and name/tempDirectoryPath of template. 
 
 First of all we define a class which handles and renders output depending on template engine used. In this case [FreeMarker](http://freemarker.incubator.apache.org/).
 
--- a/README.md
+++ b/README.md
@@ -348,7 +348,7 @@ public class JsonAcceptTypeExample {
 ```
 ---------------------------------
 
-Example showing how to render a view from a template. Note that we are using `ModelAndView` class for setting the object and name/tempDirectoryPath of template. 
+Example showing how to render a view from a template. Note that we are using `ModelAndView` class for setting the object and name/location of template. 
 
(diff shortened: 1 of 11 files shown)


tipsy:
This is a great effort @rob-signorelli, but I'm not sure if this feature is worth the one thousands lines it adds to the project. @perwendel will have to make the call.

PR #762, opened 2017-02-06
Rename from 'done' to 'afterAfter' (better name wanted)

perwendel:
Name change from 'done' do 'afterAfter' since 'done' is not really describing the functionality. We'd like to use 'finally' but it's a java keyword.

### DIFF
--- a/src/main/java/spark/Routable.java
+++ b/src/main/java/spark/Routable.java
@@ -302,7 +302,7 @@ abstract class Routable {
      *
      * @param filter The filter
      */
-    public void done(Filter filter) {
+    public void afterAfter(Filter filter) {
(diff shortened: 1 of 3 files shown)


MouettE-SC:
For the sake of completion I'd suggest renaming the DoneFilters class to AfterAfterFilters and the HttpMethod enum value from done to afterAfter. I can do a quick PR if you wish.

yuj:
if you wanted "finally" but couldn't due to reserved keyword, then how about "lastly"

tipsy:
Lastly was considered, but it just doesn't feel right:
`before`, `after`, `lastly` 🤔

ledinscak:
What about the name "atLast"

PR #765, opened 2017-02-08
Fix #714

### LINKED ISSUE #714: port() not working after initialization if port(0) is provided
This is the failing case:

```
...
port(0);
<routing setup>
....
awaitInitialization();
port() --> is returning 0, although jetty has been correctly initialised with another free port
```

Mirosta:
Calling port() after setting port(0) now returns the correct port
Joining the embedded server can be done with the join method

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -457,6 +457,11 @@ public final class Service extends Routable {
                             maxThreads,
                             minThreads,
                             threadIdleTimeoutMillis);
+                    try {
+                        server.join();
+                    } catch (InterruptedException e) {
+                        LOG.error("server interrupted", e);
+                    }
                 }).start();
             }
             initialized = true;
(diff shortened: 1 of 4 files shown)


src/main/java/spark/Service.java
```diff
@@ -453,10 +453,15 @@ public synchronized void init() {
...
                             sslStores,
-                            latch,
                             maxThreads,
                             minThreads,
                             threadIdleTimeoutMillis);
+                    try {
+                        latch.countDown();
+                        server.join();
+                    } catch (InterruptedException e) {
+                        LOG.error("server interrupted", e);
```
jakaarl: Shouldn't the interrupt flag be restored here?

Mirosta: Good point, fixed

perwendel:
:+1: 
Thanks!

Mirosta:
No problem

PR #767, opened 2017-02-09
Expose active server thread count

neenaoffline:
Is there any interest to expose data like this?

Exposing this would allow monitoring the current server threads. I'd like to use that data to fine tune the server threads configuration.

### DIFF
--- a/src/main/java/spark/embeddedserver/EmbeddedServer.java
+++ b/src/main/java/spark/embeddedserver/EmbeddedServer.java
@@ -65,4 +65,10 @@ public interface EmbeddedServer {
      * Extinguish the embedded server.
      */
     void extinguish();
+
+    /**
+     *
+     * @return The approximate number of currently active threads
+     */
+    int getActiveThreadCount();
 }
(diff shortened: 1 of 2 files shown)


src/main/java/spark/embeddedserver/jetty/EmbeddedJettyServer.java
```diff
@@ -154,5 +154,12 @@ public void extinguish() {
         logger.info("done");
     }
 
+    @Override
+    public int getActiveThreadCount() {
+        if (server != null) {
```
tipsy: If the server is NOT null, return 0?

src/main/java/spark/embeddedserver/EmbeddedServer.java
```diff
@@ -65,4 +65,10 @@ default void configureWebSockets(Map<String, WebSocketHandlerWrapper> webSocketH
      * Extinguish the embedded server.
      */
     void extinguish();
+
+    /**
+     *
+     * @return The approximate number of currently active threads
+     */
+    int getActiveThreadCount();
```
perwendel: rename to `activeThreadCount`

perwendel:
Thanks!

Issue #562, opened 2016-05-18
Client Certification Authentication


rof20004:
Hello,

I trying to auth an usb token certificate of an client, using:

secure("deploy/keystore.jks", "changeit", null, null);

and try to get with:

X509Certificate certificate = (X509Certificate)req.attribute("javax.servlet.request.X509Certificate");

On JBoss I need to setting verify-client="true" to this occur, but I do not know in Spark(Jetty).

Any help will be welcome '-'.

tipsy:
Use [stack overflow](http://stackoverflow.com/questions/tagged/spark-java) to ask general usage questions.

PR #771, opened 2017-02-12
added jetty check for client certificates

6487927a:
Address issue #562.  Just enables the jetty check to require client certificates.


chasebradford:
👍 I was just about to do this but decided to check open issues and PRs first.

perwendel:
:+1: 
Thanks!

PR #790, opened 2017-02-25
Make RouteOverview work in Java 9

stevemcleod:
I've simplified how the RouteOverview handles lambdas. It is a small step back in functionality, but  allows Spark to compile in Java 9. A positive side-effect: this commit removes the need for a build workaround that was present in pom.xml.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,10 +190,6 @@
                 <configuration>
                     <forkCount>1</forkCount>
                     <reuseForks>false</reuseForks>
-                    <excludes>
-                        <exclude>**/spark/route/RouteOverviewTest.java</exclude>
-                        <!-- Test works in IntelliJ, functionality works in jar, but this fails during mvn install -->
-                    </excludes>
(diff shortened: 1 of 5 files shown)


tipsy:
@stevemcleod I appreciate the effort, but I'm starting to wonder if it would be better to just scrap the whole route-overview functionality. Is it still useful if it just lists out all the paths? Maybe it makes more sense to just expose the routes so users can build their own route-overview?

stevemcleod:
> Maybe it makes more sense to just expose the routes so users can build their own route-overview?

That makes sense to me.

I don't mind getting rid of the functionality altogether. It was a good exercise for me to learn a bit more about Spark.

tipsy:
> I don't mind getting rid of the functionality altogether. Trying to fix it for Java 9 was a good exercise for me to learn a bit more about Spark.

Ok, great, I'll close this PR. Feel free to make a new one removing all RouteOverview functionality.

PR #792, opened 2017-02-28
Optimized UriPath.java and created UriPath.test (formatted)

tipsy:


tipsy:
Just tested this, and it's about 4 times slower than the current implementation. I will add the test though.

PR #793, opened 2017-03-01
Removed all RouteOverview functionality

stevemcleod:
RouteOverview functionality wouldn't compile because it used an internal Sun API not available in Java 9. The existing functionality also seemed to vary in results depending on Java environment.


tipsy:
Great, thanks!

stevemcleod:
I just noticed that the documentation here refers to the RouteOverview:
 http://sparkjava.com/documentation.html

I think I can't change the docs. @perwendel could you please remove the RouteOverview section from that page?

tipsy:
Well, it's still in there for the current versions, so I think it's fine to keep it in the docs. I will remove it from the docs when I launch the new website.

PR #794, opened 2017-03-01
Made minor changes to avoid compiler warnings

stevemcleod:
* Removed unused import that caused compiler warning
* Replaced use of deprecated URLEncoder.encode(String s) method with URLEncoder.encode(String s, String enc) alternative.


tipsy:
Thanks!

PR #800, opened 2017-03-09
Adding a method to configure maximum allowed http headers size

spuklo:
I've stumbled upon an issue that I've had to deal with client that is using huge headers (more than 16k) and Jetty's defaults were not good enough. This PR provides a Spark.maxHeadersSize(size) method that sets max headers size for both requests and responses.


tipsy:
This should be taken care of by #813

PR #801, opened 2017-03-09
Implemented default value for queryParams in Request

orby:
A real killer feature :-)

### DIFF
--- a/src/main/java/spark/Request.java
+++ b/src/main/java/spark/Request.java
@@ -16,6 +16,14 @@
  */
 package spark;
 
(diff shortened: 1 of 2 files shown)


tipsy:
Takk! I already had the same feature in a branch locally, so I pushed that. (8cbf214909b509f9bba9738a0e3ae1274f8c9364)
My method name is `queryParamOrDefault`, which I think fits better since `getOrDefault` is an established concept in Java, and `methodName(s1, s2)` looks a lot like a setter in Spark.

Issue #739, opened 2017-01-03
externalStaticFileLocation forces browser downloads of files since 2.5.4


toni07:
Hi, I previously used the 2.5 version with this code: 
`externalStaticFileLocation("src/main/resources/public/");`

I have an "index.htm" file in src/main/resources/public/, and with the 2.5 version I could see my html page in a browser. When I switch to the 2.5.4 version, the browser forces me to download the html page, thus I can no longer see it.

I did not check the header of the HTTP response, but I suppose it is something like "Content-Description: File Transfer".

Cheers and thanks for this great framework!

tipsy:
Maybe not related, but `src/main/resources` should be `staticFiles.location()` not  `staticFiles.externalLocation()` ?

toni07:
This is the same problem with staticFiles.location().
The header of the server response (content-type) is: "application/octet-stream".

PR #808, opened 2017-03-26
Add mapping for ".htm" files

toni07:
For these old school guys like me who like 3 letters extensions files.
Fixes issue #739.


tipsy:
Thanks!

PR #819, opened 2017-04-08
Revert "Make RouteImpl factory methods public"

tipsy:
Reverts perwendel/spark#692


tipsy:
Not the problem it seems.

PR #821, opened 2017-04-08
Disable stream cache for chunked transfer encoding

tipsy:


tipsy:
> The current implementation for the HttpServletRequest returned from request.raw() (HttpRequestWrapper.class) caches the bytes when getInputStream() is called. When large data is transfered, an OutOfMemoryError is thrown, even when using chunked transfer encoding.

PR #834, opened 2017-04-17
Fix CRLF in CustomErrorPages.java

MouettE-SC:
CustomErrorPages.java has been commited with CRLF line endings in the repository whereas all files are LF-terminated


tipsy:
Thanks, that explains it.

PR #813, opened 2017-03-30
Make Jetty servers tunable

gquintana:
Fixes #796  #314
Adds an extension to let people configure their Jetty Server

PR #836, opened 2017-04-20
Expose Jetty Session Clustering

kendrickwilson:
updated to add session clustering


feg624:
Hi, by adopting the changes from a recent pull request to configure Jetty (#813), you can use Jetty's default capabilities to have clustered sessions. In my case, I was able to setup Jetty's nosql session management, which relies on MongoDB.

My current blocker is that I cannot close the open connection to MongoDB once the Spark app/microservice is stopped.

I thought this comment would be useful for this PR.

Franco

kendrickwilson:
I would like to see more information in regard to this. Do you have a project on github?


If this can be done with standard jetty, it may be the preferred way to do it. I did notice jetty stores a binary version of the session data object into the database. I would think a json/text object is a better approach.


Kendrick C. Wilson


________________________________

feg624:
I don't have a sample code on Github, as I'm still doing some investigation development to see whether I use Spark for microservices development.

I basically implemented my own class, which is a copy of spark.embeddedserver.jetty.JettyServer . This class is actually the one responsible of creating the Jetty instance. Within my own class I provided the sessionIdManager and sessionHandler to the Jetty server instance.

So, within the main method of the app (prior to making calls to other Spark methods), I've used the following code to setup my own class (MyJettyServer) instead of JettyServer:

EmbeddedServers.add(EmbeddedServers.defaultIdentifier(), new EmbeddedJettyFactory(MyJettyServer::create));

Within my own class, I hace direct access to Jetty server instance, so in there I can configure the clustered session with MongoDB, or other SQL approach supported by Jetty, and any other custom configuration.

Regarding not releasing resources, I found that it's normal for Java Mongo client to open two connections, and those are actually closed once the Spark app finishes. Anyways, I still think that having some place to manually close the resources would be safer and more elegant.

Regards,

Franco

kendrickwilson:
Closing pull request. Do to changes in jetty 9.4.4....

PR #698, opened 2016-10-31
Updated Jetty dependency and a test broken by Jetty changes.

jakaarl:
See issue: #694

PR #838, opened 2017-04-23
Issue #698 - Upgrade to Jetty 9.4.4

joakime:
Here's an upgrade to the recently release Jetty 9.4.4.v20170414


src/test/java/spark/staticfiles/StaticFilesTest.java
```diff
@@ -142,8 +143,13 @@ public void testDirectoryTraversalProtectionLocal() throws Exception {
         String path = "/" + URLEncoder.encode("..\\spark\\", "UTF-8") + "Spark.class";
         SparkTestUtil.UrlResponse response = doGet(path);
 
-        Assert.assertEquals(404, response.status);
-        Assert.assertEquals(NOT_FOUND_BRO, response.body);
+        // Attempt to access context above root is either a 400 or 404 depending on environment
```
tipsy: This change shouldn't be necessary if you rebase your patchset. I already changed the behavior of the path-traversal guard to return 400 instead of 404.

src/test/java/spark/InitExceptionHandlerTest.java
```diff
@@ -16,18 +16,16 @@
...
     public static void setUpClass() throws Exception {
-
         service1 = ignite();
         service1.port(1122);
         service1.init();
         service1.awaitInitialization();
 
         service2 = ignite();
-        service2.port(1122);
+        service2.port(70000);
```
tipsy: Not related to change? I suppose it's a better way of testing it than to fire up two instances though, I will update the test.

tipsy:
Thanks @joakime ! 
I fixed the test you adjusted in your commit. If you rebase you can get rid of that + the static-files-test change you made.

tipsy:
Thanks a ton!

src/test/java/spark/InitExceptionHandlerTest.java
joakime: Starting 2 servers on the same port is supported in some environments.
This test fails on my environment, as its allowed to have 2 servers on the same listening port.

tipsy: Makes sense. I included this change in 8706ba4 before merging.

src/test/java/spark/staticfiles/StaticFilesTest.java
joakime: This is a general truth about security on Jetty with regards to path traversal and OS and/or FileSystem specific alias behavior.

You *really* don't want to be checking paths on your own, its a *very* large subject and has some nasty corner cases.

Some examples path traversal testing in Jetty ...

How it works ..

1. Take the provided Request URI and perform a canonicalPath() resolution to attempt to determine the contextPath to send the request to
2. Send request to specific context, with raw, as sent, request URI as the pathInfo
3. The specific url-pattern within that context is ascertained from the raw request URI (not processed through canonicalPath)
4. If the request doesn't hit a Servlet, then it goes to the default path (aka `<url-pattern>/<url-pattern>`)
5. In a normal Servlet Context, the "Default Servlet" (concept from the Servlet Spec) is hit to determine if this is a static file, or an error (such as 404)
6. The jetty `DefaultServlet` will perform a `canonicalPath(requestURI)` to determine what content with the "Base Resource" location is being requested.
7. If the requested path exists, a check to see if the requested path is actually an alias to the real path is performed (to ensure that the vast variety of security rules are still performed against the real path)

https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-util/src/test/java/org/eclipse/jetty/util/URIUtilCanonicalPathTest.java

If the expectedResult is null, this is a context-less request and will result in an error 400 (no context is asked for its error specifics)

If the expectedResults start with a "/" then this is valid for use in url-pattern / servlet / filter lookups.

If the expectedResults start with anything else, then this is considered relative to the Servlet Base Resource for static file lookups.   Those lookups go through extra security steps to ensure that they belong within the Base Resource only (and not outside of it)

Some examples of tricky alias lookups.

```
-- file allowed on unix, but not windows
/dir?/index.html  (windows does not allow "?" in file/dir name)
"file://Z:/:/" (windows only allows 1 colon)
"file:://CON"  (windows reserved words)
"file' name" (invalid name on windows)
-- filenames that are the same on windows
"TextFile.Long.txt"  == "TEXTFI~1.TXT" (8dot3 alias)
"testfile" == "testfile:stream" (ntfs stream)
"testfile" == "testfile:$DATA" (ntfs data stream)
"testilfe" == "testfile::%24DATA" (ntfs data stream, UTF8 encoded at filesystem level)
"testfile" == "testfile    " (extra spaces at end ignored)
"testfile" == "testfile....." (extra dots at end ignored)
"testfile" == "testfile\r" (extra CR ignored at end)
"testifle" == "testfile\u0000" (null ignored at end)
"testfile" == "testfile\"\"\"" (extra quotes ignored at end)
"testfile" == "testfile<<<<>><>>><" (angle brackets ignored at end)
"testfile" == "testfile;" (extra semi-colons ignored at end)
```

You can see these, and a few more with the FileSystemTests at Jetty
https://github.com/eclipse/jetty.project/blob/jetty-9.4.x/jetty-util/src/test/java/org/eclipse/jetty/util/resource/FileSystemResourceTest.java

tipsy: I fully agree. Could you create an issue from your comment so everyone can see it?

PR #843, opened 2017-04-24
Added ServletContext parameter to SparkApplication.init

MouettE-SC:
This is a small change which doesn't break the filter API but allows its users to access to the servletcontext while initializing if they wish.


MouettE-SC:
Pushing a little further we could also add a @Deprecated annotation to the original init() call

oeph:
@perwendel Is there any chance for this to get merged?

PR #849, opened 2017-05-02
Add configurable path parameter matcher

hydroid7:
Now you can set the regular expression with `SparkUtils.paramMatcher(regex);`
There are two default values:
1) `DEFAULT_PATTERN` which matches the old style
2) `CURLY_BRACKET_MATCHER` which matches the curly bracket style used by Spring and OkHttp
3) You can also set your own pattern with the `paramMatcher(regex)` method.

### DIFF
--- a/src/main/java/spark/utils/SparkUtils.java
+++ b/src/main/java/spark/utils/SparkUtils.java
@@ -27,8 +27,27 @@ import java.util.List;
 public final class SparkUtils {
 
     public static final String ALL_PATHS = "+/*paths";
+    /**
+     * The default value for the path matcher and follows the standard convention.
+     * Path parameters can be defined as {@code :parameter}.
+     */
+    public static final String DEFAULT_PATTERN = ":.+";
+
+    /**
+     * Matcher for the curly bracket notation. This pattern is used by many
+     * popular frameworks like Spring MVC or OkHttp. The expression matches
+     * strings like {@code {id}, {value}} or {@code {pathVariable}}.
+     */
+    public static final String CURLY_BRACKET_MATCHER = "\\{.+\\}";
+
+    private static String regex = DEFAULT_PATTERN;
 
     private SparkUtils() {
+    }
+
+
+    public static void setParamPlaceholder(String regex) {
+        SparkUtils.regex = regex;
     }
 
     public static List<String> convertRouteToList(String route) {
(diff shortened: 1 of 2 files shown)


jakaarl:
The changes seem straightforward, but...

I don't see the point - you can't use Spark as a drop-in replacement for Spring or OkHttp, besides technical differences it's really a case of apples and oranges. What need does this serve, other than presenting the developer with yet another choice? Also, this is exposing mutable state via a static method, which is generally not a good idea.

IMHO, 1) should evaluate if this is really necessary and 2) if yes, refactor the path matching so that it doesn't have static mutable state.

Just my opinion, everyone is welcome to convince me otherwise.

hydroid7:
As you mentioned, of course the two notations are just like apples and oranges. 

The obvious question is why I needed these changes?

In my case I've written a REST API with spark. The client is a rich Java based client with Retrofit.
I declared a class that holds all route addresses:
```
public final class Routes {
    private Routes () { }
    public final static String ALL_USER = "user/all";
    public final static String USER = "user/:id";
    ...
```
Then I use these to map the requests in Spark:
```
Spark.get(Routes.ALL_USER, (request, response) -> { ... }
```
I also use `Routes` in my http client.
```
@GET(Routes.ALL_USER)
Call<List<User>> listAllUser();
```
In Retrofit you can pass parameters to your query, like this:
```
@GET(Routes.USER + "/{id}")
 Call<User> userById(@Path("id") long id);
```
But for the Spark mapping, I code as follows:
```
Spark.get(Routes.USER + "/:id", (req, res) -> { }
```

But if I could configure the matcher, I could user the same mapping for server and client. Of course it is not recommended to make static state, but I think this part is very simply and can be easy understand.

tipsy:
I'm not a big fan of our use of statics, but it seems like this would be a very global setting (you would never want to launch two instances with different param matchers?). I'm not sure if this is needed though.

jakaarl:
@hydroid7 sounds like a valid use case. I wonder if path parameter "matching mode" could be set on a `Service` instance and get _somehow_ passed along to the code doing the matching. From a quick look, this would require some refactoring, though.

hydroid7:
Do you think about a `PathPreprocessor` object which can be set on a `Service` object?

perwendel:
Update:

perwendel:
Thanks for contributing.

However, I'm not a fan of having multiple choices for doing the same thing! There should be one way of declaring path parameters.

arun-y:
@hydroid7 - I guess the best way to contribute via PR is to first gets an agreement with maintainers. This may save time by avoiding real implementation. @perwendel @tipsy - Do we have any contribution guidelines? may be something in line with https://help.github.com/articles/setting-guidelines-for-repository-contributors/

Issue #297, opened 2015-06-13
Can we have a non-static API?


gubby:
(Perhaps there is already a way to do this, that I have missed... but looking at spark.SparkBase it looks unlikely)

The fact that all the methods (e.g. port(), get(), etc) are static, means that code using Spark is impossible to test, and Spark is no use in integration tests. Additionally, it prevents two servers being started in the same JVM (e.g. HTTPS/HTTP), which makes it unsuitable for quite a large number of use cases (e.g. HTTP-to-HTTPS redirection, for which I still need to run e.g. Apache).

Would it be possible to expose a more traditional API that is based on instance methods? I don't think the fluidity would be lost substantially:

```
Spark spark = new Spark();
spark.port(8080);
spark.get("/", this::homePage);
etc.
```

Thanks!

phase:
You could copy all the methods and remove the `static`.

gubby:
@phase Everything in spark.SparkBase would also need changing, and in a forked repo.

magicgoose:
+1, this stuff needs refactoring, static API is inherently less flexible

amarseillan:
im trying to implement this, but it gets hard when embedded in another application server, will keep you up to date!

ribomation:
*Suggestion*: let the methods return `this`, aka cascade calls. 
For example:
```java
    new Spark()
      .port(8888)
      .get("/",...)
      .get("/ping",...)
      .start();
```

magicgoose:
> Suggestion: let the methods return this, aka cascade calls. 

there's little benefit anywhere except "hello world" server examples…

amarseillan:
>Suggestion: let the methods return this, aka cascade calls.
>there's little benefit anywhere except "hello world" server examples…

Yeah... unless you want it to be immutable for thread safety?

magicgoose:
I don't see how it's related to immutability.
Anyway, I'm not saying that I'm against it, it'll probably be good to have

amarseillan:
> I don't see how it's related to immutability.
You copy-on-write and return the new instance, just that

amarseillan:
So I think I'll just upload a version without the feature of deploying spark in different application servers because I don't use that and I can't find a generic way to achieve it without static properties, I'll post the repo here when I have that so someone can check it out and maybe fix that

magicgoose:
So it'll be possible to run 2 or more servers in one Java application? Then it's cool!

amarseillan:
Yup =D!

[...]

PR #858, opened 2017-05-18
Restoration of RouteOverview Functionality

ericfialkowski:
We found the RouteOverview very useful and were sad to see it removed in 2.6. This PR adds it back without the internal sun API which required a small change in the output.

### DIFF
--- a/src/main/java/spark/route/RouteOverview.java
+++ b/src/main/java/spark/route/RouteOverview.java
@@ -1,0 +1,153 @@
+/*
+ * Copyright 2011- Per Wendel
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package spark.route;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


tipsy:
Thanks Eric, I'll have a look when I get back from vacation. Is there any change in behaviour?

ericfialkowski:
The output for lambdas/method references changed. I haven't found a way to get the previous data without the internal sun classes.

StuAtGit:
One *minor* nice-to-have for this PR would be to also enable routeOverview for Service instances.

This issue:
https://github.com/perwendel/spark/issues/297

in theory implemented the spark static interface on an instance - however, it doesn't support routeOverview(). I can file as a separate ticket, if that works better.

Note that routeOverview is a must-have feature for us (at work, not just in my play programs), in order to properly support, at very least, bare-bones, almost-hateoas/restful services:
https://en.wikipedia.org/wiki/HATEOAS

src/main/java/spark/route/RouteOverview.java
```diff
@@ -0,0 +1,94 @@
...
+    static void add(RouteEntry entry, Object wrapped) {
+
+        if (wrapped instanceof Wrapper) {
+            entry.target = ((Wrapper) wrapped).delegate();
+        }
+
+        routes.add(entry);
+    }
+
+    static String createHtmlOverview(Request request, Response response) {
```
StuAtGit: Would love if this was public, so it can be used with a spark instance, as outlined in my comments.

StuAtGit: Would love it if this was public so it can be used with instances like so:
```
Service sparkService = Service.ignite();
sparkService.get("/api", RouteOverview::createHtmlOverview);
```
instances are outlined in:
https://github.com/perwendel/spark/issues/297

ericfialkowski: It's an easy enough change. Should it be a separate change from this PR, since it is a further change from what is trying to be restored?

tipsy:
@ericfialkowski It's fine to add the functionality to the Service class in this PR.

perwendel:
Update:

PR #860, opened 2017-05-19
don't do route.render if response is already redirected

### LINKED ISSUE #859: Don't do route.render when response.isRedirected() is true
Sometimes we need do redirect or render a TemplateView ,depends on some conditions , like this:
`if (someCondition) {
    response.redirect("some/other/url");
    return null; // we 'return null' here because we don't want to prepare the context for render the view
} else {
   // prepare context for the view
   return new ModelAndView(xxxx);
}
`

in this case ,spark still try to render the ModelAndView(it's null now) , so it throws NPE.

argan:
fix #859


tipsy:
Thanks!

PR #874, opened 2017-06-15
Enable custom 404 pages for external web containers

mlitcher:
Removed external container check. This gives external web containers the ability to use CustomErrorMessages.

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -532,6 +532,16 @@ public final class Service extends Routable {
         }
     }
 
+    /**
+     * @return The approximate number of currently active threads in the embedded Jetty server
+     */
+    public synchronized int activeThreadCount() {
+        if (server != null) {
+            return server.activeThreadCount();
+        }
+        return 0;
+    }
+
     //////////////////////////////////////////////////
     // EXCEPTION mapper
     //////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


tipsy:
Since this behavior is very old (pre custom error pages), I want to check with @perwendel before merging it.

perwendel:
I'm trying to figure out why `&& !externalContainer` is there to begin with!

mlitcher:
Since this is the only usage of the `externalContainer` flag, would it make sense to also remove it completely from this class, including the constructor? If you agree, I can update this PR.

perwendel:
Update:

src/main/java/spark/Service.java
```diff
@@ -532,6 +532,16 @@ private void initializeRouteMatcher() {
         }
     }
 
+    /**
+     * @return The approximate number of currently active threads in the embedded Jetty server
+     */
+    public synchronized int activeThreadCount() {
```
perwendel: rename to `activeThreads()`

src/main/java/spark/Spark.java
```diff
@@ -1197,4 +1197,11 @@ public static ModelAndView modelAndView(Object model, String viewName) {
         return new ModelAndView(model, viewName);
     }
 
+    /**
+     * @return The approximate number of currently active threads in the embedded Jetty server
+     */
+    public static int activeThreadCount() {
```
perwendel: rename to `activeThreads()`

mlitcher:
Just looking back at this, it looks like I inadvertently pushed unrelated commits into this pull request. I'm thinking I should re-do this pull with only my change?

tipsy:
> I'm thinking I should re-do this pull with only my change?

Please do.

PR #902, opened 2017-08-15
Use `StaticFilesConfiguration` as the single source of truth.

jakaarl:
Ditch local variables in `Service` in favour of consulting
the configuration object. Also fixes the issue of not properly
clearing on stop.


tipsy:
This looks very nice, thanks.

Issue #920, opened 2017-09-26
Awaiting server intialization does not succeed


ORESoftware:
I have the following code:

```java
package cdt;
import org.apache.log4j.*;
import spark.Spark;
import java.sql.*;

public class App {

    public static void main(String[] args) throws ClassNotFoundException, SQLException {

        int maxThreads = 8;
        int minThreads = 2;
        int timeOutMillis = 30000;
        Spark.threadPool(maxThreads, minThreads, timeOutMillis);

        System.out.println("began initialization.");
        Spark.awaitInitialization();
        System.out.println("finished initalizing");
        Spark.get("/hello", (req, resp) -> "Hello World");

    }

}
```

the problem is "finished initializing" is never logged. Why? :(

ORESoftware:
If I remove the `Spark.awaitInitialization();` call, I get this error:

```console
[WARNING] 
java.lang.NoSuchMethodError: org.eclipse.jetty.util.thread.QueuedThreadPool.<init>(III)V
        at spark.embeddedserver.jetty.JettyServer.create(JettyServer.java:43)
        at spark.embeddedserver.jetty.EmbeddedJettyServer.ignite(EmbeddedJettyServer.java:91)
        at spark.Service.lambda$init$0(Service.java:349)
        at java.lang.Thread.run(Thread.java:745)

```

tipsy:
`awaitInitialization();` blocks the thread until the server has started, so you need to call it after `get("/hello", (req, resp) -> "Hello World");` (this is the method call which initializes the server).

It's in the docs: http://sparkjava.com/documentation#awaitinit

the1mills:
If no routes have been registered yall need to throw an error, otherwise bad dev experience.

ORESoftware:
a call to get() should flip a boolean on Spark object. if the boolean has not been set, and awaitInitialization(); is called, then please throw an err people, it helps us newbs.

tipsy:
If one of you can make a PR, I'll merge it.

ORESoftware:
I am so out of the loop in the Java world, I haven't coded Java in years

mlitcher:
I submitted a pull for this. There is already a boolean which tracks server initialization which can be tapped into.

ORESoftware:
@mlitcher thanks

PR #923, opened 2017-09-28
Add initialization check before blocking

mlitcher:
Resolves issue #920. Not sure if there are better suggestions on the exception message?


tipsy:
It's fine, I think. Thanks!

PR #927, opened 2017-10-08
removed unreachable if conditions

mirkoperillo:
Using findbugs found two inner if blocks explicitly always true or false


perwendel:
:+1: 
Thanks!

PR #951, opened 2017-12-13
Replaced explicit null-checks by a declarative approach

denis-zhdanov:
The idea is to do the following:  
1. Remove explicit *null*-checks
2. Mark corresponding method parameters by *Nonnull* annotation
3. Configure the project to use [Traute](http://traute.oss.harmonysoft.tech/) *javac* plugin to enhance resulting bytecode by *null*-checks for the target method parameters  

Example: consider [ClassPathResource](https://github.com/denis-zhdanov/spark/commit/b9159ecbf8e45b2740d937ae7fbf5e5222cb25bf#diff-30e3cccc077c04f39eb783f91945edd7):  

**Before**  

```java
public ClassPathResource(String path, ClassLoader classLoader) {
    Assert.notNull(path, "Path must not be null");
    String pathToUse = StringUtils.cleanPath(path);
    ....
```  

**After**  

```java
public ClassPathResource(@Nonnull String path, ClassLoader classLoader) {
    String pathToUse = StringUtils.cleanPath(path);
    ....
```  

Resulting bytecode:  

```
javap -c ./target/classes/spark/resource/ClassPathResource.class
...
  public spark.resource.ClassPathResource(java.lang.String, java.lang.ClassLoader);
    Code:
       0: aload_0
       1: invokespecial #2                  // Method spark/resource/AbstractFileResolvingResource."<init>":()V
       4: aload_1
       5: ifnonnull     18
       8: new           #3                  // class java/lang/IllegalArgumentException
      11: dup
      12: ldc           #4                  // String 'path' must not be null
      14: invokespecial #5                  // Method java/lang/IllegalArgumentException."<init>":(Ljava/lang/String;)V
      17: athrow
```  

Benefits:  
* the code is cleaner because explicit *null*-checks are removed
* the code is better documented because it clearly indicates which parameters can't be *null*
* IDEs highlight possible *NPE* problems for method parameters marked by *Nonnull*

TESTED: mvn clean package & ensured
        that resulting bytecode contains null-checks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,6 +75,19 @@
             <version>${jetty.version}</version>
         </dependency>
 
+        <!-- NULL-CHECK DEPENDENCIES -->
+        <dependency>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+            <version>3.0.2</version>
+        </dependency>
+        <dependency>
+            <groupId>tech.harmonysoft</groupId>
+            <artifactId>traute-javac</artifactId>
+            <version>1.0.10</version>
+            <scope>provided</scope><!-- make the jar eligible for compilation only -->
+        </dependency>
+
         <!-- JUNIT DEPENDENCY FOR TESTING -->
         <dependency>
             <groupId>junit</groupId>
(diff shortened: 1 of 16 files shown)


tipsy:
We meet again @denis-zhdanov. I'm not sure we want to introduce this in 2.X. @perwendel can decide.

denis-zhdanov:
Hi again :)

Any particular concern about that?

tipsy:
Not really any concerns, but the current version is supposed to be the last 2.X, with the exception of bugfixes.

denis-zhdanov:
Ah, fair enough.
I don't think it's necessary to make a dedicated release for that, i.e. it's enough to apply the feature and then release when there is a significant feature/bugfix to deliver to end-users.

denis-zhdanov:
Hey @perwendel  

Could you please share your thoughts on the idea of adding *Traute* to your project?

Regards, Denis

perwendel:
Thanks! Currently there are much higher prioritized bugs/features that needs attention. Adding another lib for null checks does not feel important, even though I like how it seems to work.

PR #971, opened 2018-01-28
Fix #506

### LINKED ISSUE #506: Add SslStores & Spark::secure support for a specific keystore alias
Generally I work with more than one KeyPair in keystores, and for that it would be a good additional benefit to be able to select a KeyPair by its alias name in the #secure call (otherwise just the first valid KeyPair would selected from the given keystore).

The SslStores class + the #secure method in the Spark class could be improved concerning that.

AFAIK the JVM itself does not support this in the corresponding VM parameters (java.net.ssl. ...), so it's up to the application to handle KeyStores with multiple KeyPairs.
- For example Jetty supports such an alias by a Property "certAlias" in its org.eclipse.jetty.util.ssl.SslContextFactory.
- Other products use other name (certAlias, keyAlias, javax.net.ssl.keyStoreAlias) or rely on the first certificate in the KeyStore.

If this is acceptable for you, I could provide a pull request for an appropriate solution.

If not we can discuss this in detail.

rozhko:
Add support for a specific keystore alias.

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -180,7 +180,32 @@ public final class Service extends Routable {
                                        String keystorePassword,
                                        String truststoreFile,
                                        String truststorePassword) {
-        return secure(keystoreFile, keystorePassword, truststoreFile, truststorePassword, false);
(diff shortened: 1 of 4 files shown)


perwendel:
:+1:

PR #978, opened 2018-02-22
Ensure InputStreams returned by Routes are closed

thomaslee:
Howdy -- we ran into a situation where resources could be leaked when returning `InputStream`s from `Route.handle()`. Proposed patch below, happy to hook up tests if the approach looks reasonable.


perwendel:
Thanks for commiting. I've myself never returned InputStreams from a Route. Is it conceivable that the serializer handling the inputstream would be responsible for closing the stream?

ThatJoeMoore:
We ran into this today, so I created #1121, which incorporates this feedback.

PR #983, opened 2018-02-26
Add support for asynchronous request handling

joatmon:
See comment in: Support of event based (non-blocking) request processing. #549

Duplicated here:
Hi,

My team has been using Spark as a platform to build microservices for the last year. We have one service that is still running on Spring Boot that we would like to port to Spark, but it makes heavy use of Asynchronous Requests. We've reviewed the PR from mj1618, and don't think that it would work for our use case because we also need the ability to stream back very very large responses.

So we have come up with an alternative PR that we have tested with our application (at transaction rates of up to 3000 requests per second)

# TL;DR

This PR:
* adds support for Asyncronous Request handing using *any* Java construct you wish (threads, futures, ...)
* has no impact on existing Spark applications
* has a minimal impact on the Spark code base
* supports *both* Spark style request API's and traditional Servlet 3.0 style API's
* supports Spark After and AfterAfter filters.

# USE CASE 

I'll use our service as an example of why asyncronous request handling is needed. Our service basically exposes a set of REST API's that can be used to issue queries against any of several backend databases. The user of the REST API can submit any query they wish, so we have no idea how long the query will run or how much data it will return. It could run for 1ms and return 1 row of data or it could run for 2 hours and return a million rows of data.

In typical syncronous request processing, these queries would be running in the Jetty request thread. So if enough long running queries are submitted, we will run out of request threads. 🙁 

So instead we use asyncronous requests which free up the Jetty request thread once we have put the request on an internal queue. Then we have our own thread pool that executes database queries from the internal queue and then streams the results back to the client.

Because we are potentially returning so much data, we cannot keep  all of the response body in memory at one time. So we need to be able to stream the response directly to the output stream of the HTTP response. This is the part that doesn't work well with the existing Spark API's because they store the response body in string.

# EXAMPLE: Spark API's

Here is a simple example of a Spark Controller method that creates its own thread to use to process the request asyncronously:

```
        get("/async", (req, res) -> {
            final AsyncContext ctx = req.startAsync();
            Thread t = new Thread(() -> {
                try {
                    Thread.sleep(500);
                    res.body("Hello from Async!");
                    ctx.complete();
                }
                catch (Exception e) {
                    e.printStackTrace();
                }
            });
            t.start();
            return "";
        });
```

# EXAMPLE: Servlet 3.0 API's

Here is the same example implemented with Servlet 3.0 API's. 

```
        get("/traditional", (req, res) -> {
            final AsyncContext ctx = req.raw().startAsync();
            Thread t = new Thread(() -> {
                try {
                    Thread.sleep(500);
                    PrintWriter writer = ctx.getResponse().getWriter();
                    writer.write("lots of data");
                    writer.close();
                    ctx.complete();
                }
                catch (Exception e) {
                    e.printStackTrace();
                }
            });
            t.start();
            return "";
        });
```

Both of these approaches are supported by this PR.

### DIFF
--- a/src/main/java/spark/Request.java
+++ b/src/main/java/spark/Request.java
@@ -28,10 +28,12 @@ import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
 
+import javax.servlet.AsyncContext;
 import javax.servlet.http.Cookie;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
 
+import spark.http.matching.AsyncContextWrapper;
 import spark.routematch.RouteMatch;
 import spark.utils.urldecoding.UrlDecode;
 import spark.utils.IOUtils;
@@ -490,6 +492,15 @@ public class Request {
      */
     public String protocol() {
         return servletRequest.getProtocol();
+    }
+
+    /**
+     * Starts an asynchronous request.
+     *
+     * @return the asynchronous context
+     */
+    public AsyncContext startAsync() {
+        throw new RuntimeException("this should be not be called; use RequestWrapper.startAsync instead");
     }
 
     private static Map<String, String> getParams(List<String> request, List<String> matched) {
--- a/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyFactory.java
+++ b/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyFactory.java
@@ -16,7 +16,9 @@
  */
 package spark.embeddedserver.jetty;
 
+import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.util.thread.ThreadPool;
+
 import spark.embeddedserver.EmbeddedServer;
 import spark.embeddedserver.EmbeddedServerFactory;
 import spark.http.matching.MatcherFilter;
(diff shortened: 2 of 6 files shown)


joatmon:
In the benchmark above I implemented an even cleaner way of presenting async request handing to the sparkjava application developer. With this approach developers write async request handlers exactly the same way that they do normal request handlers. They just use a different API to register the handler. Here is example code from the benchmark. Note that I am able to use the exact same method to process requests synchronously or asynchronously.

```java
public class DataController {
    DataService service;

    public DataController(DataService service) {
        this.service = service;
        Spark.get("/data", this::getData);
        SparkAsync.getAsync("/asyncdata", this::getData);
    }

    private String getData(Request req, Response resp) throws Exception {
        long delay = 0;
        String delayParam = req.queryParams("delay");
        if (delayParam != null) {
            delay = Long.parseLong(delayParam);
        }
        return service.getData(delay);
    }
}
```

danforbes:
@perwendel - can you please let us know your thoughts on this pull request? We believe this is a very useful feature and it has been thoroughly tested by my team as we move our Spring Boot microservice onto the excellent Spark framework. We are happy to work with you to iron out any issues you see.

perwendel:
@danforbes sorry for not addressing this earlier. We have limited bandwidth. I will review this soon!

PR #998, opened 2018-04-10
Use a default response transformer

richarddd:
I would find it useful to add a default response transformer. I'm using spark as a REST API thus all my responses should be json.

`setDefaultResponseTransformer(new JsonTransformer());`

_I also removed some unused imports from `Service.java` (and applied formatting)_

### DIFF
--- a/src/main/java/spark/Routable.java
+++ b/src/main/java/spark/Routable.java
@@ -24,6 +24,8 @@ import spark.utils.SparkUtils;
  */
 abstract class Routable {
 
+    private ResponseTransformer defaultResponseTransformer;
+
     /**
      * Adds a route
      *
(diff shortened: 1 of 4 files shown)


richarddd:
@tipsy @flyrev could you please look at this? 👍 :)

tipsy:
Nice, looks good to me. As a general rule you should try to keep "general cleanup" out of feature commits, so that the PR is easier to review. This is a pretty significant API change, so @perwendel will have to approve it.

richarddd:
@tipsy I've updated the PR to remove formatting. Turns out my IntelliJ config applied auto-formatting on save which used star imports and destroyed the default formatting.

@perwendel Please check this PR when you have a few minutes over. Would significantly remove a lot of boilerplate in my project :)

richarddd:
@tipsy and @perwendel Understand that you are super busy but this should not be a major change :) Please check as soon as you have time!

tipsy:
Looks good to me, but Per wants to merge API changes himself.

perwendel:
@richarddd @tipsy 
reviewing!

Issue #1013, opened 2018-05-18
Service::stop() does not allow me to know when the job  is done


chibenwa:
Hi everyone.

One of my collegues recently wrote a test about default exception handling, involving registration of exception handler. Curiously, test order impacted test correctness (breaking test isolation) leading to a funny debug session.

It turned out the `Service::stop` is run in an other thread. Thus our test suite was starting a new test before the `stop` operation finished. A data race between the new test and the old test resulted in the handler of the new test being  deleted by the old test....

Needless to say, I need to achieve strong test isolation. My proposal is simple: return  Future that the user can await on **to ensure  cleaning process finished**.

I will have a pull request for this.

simonreye:
I hit this problem today as well.

May I suggest we keep stop() as is and have an additional function stopAndWait() (or something like that) for those who don't want the stopping calls to be done in a separate thread.

jakaarl:
See PR #730 .

iamybj:
this framework is too naive.

redcatbear:
:+1: for addressing this.
Ran into this problem with automatic integration tests where I have to start and stop regularly. My current workaround using a wait() is not very satisfying.

chibenwa:
Yes that is adresses.

PR #1014, opened 2018-05-18
ISSUE-1013 Service::stop() should allow to await completion

chibenwa:
This would allow to gain control on that operation and help avoid data races.

Solves issue https://github.com/perwendel/spark/issues/1013


jakaarl:
Related to: #730 .

perwendel:
#730 has been merged. I suppose that makes this redundant?

chibenwa:
For sure!

PR #1017, opened 2018-05-24
Refactors

ErnestoSalazar:


perwendel:
Thanks for contributing but the minor changes here contains format errors etc that in turn would need new refactoring. Also, since it's only in test classes it's not worth merging since it'll create merging needs for other "more important" PRs which costs time we dont' have.

PR #1027, opened 2018-06-27
Set HttpOnly for session cookie due to improvement of security

asolntsev:
This is an improvement for PR https://github.com/perwendel/spark/pull/965 from @M-Razavi

In this PR, I
1. I made it possible to disable `HttpOnly` flag (just in case, probably somebody will need it).
2. Added unit-tests.

### DIFF
--- a/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyFactory.java
+++ b/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyFactory.java
@@ -29,6 +29,7 @@ import spark.staticfiles.StaticFilesConfiguration;
 public class EmbeddedJettyFactory implements EmbeddedServerFactory {
     private final JettyServerFactory serverFactory;
     private ThreadPool threadPool;
+    private boolean httpOnly = true;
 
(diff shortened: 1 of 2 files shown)


asolntsev:
@perwendel @tipsy @jakaarl @joatmon @M-Razavi ping

asolntsev:
@perwendel PING!

asolntsev:
@perwendel Hello! Is anybody alive?

perwendel:
@asolntsev 
Hey, two questions:
* does this change any default behavior? Is httpOnly 'true' by default? (I suppose it's the case)
* how does this relate to the `Response.cookie` where `httpOnly` can be set on individual cookies?

asolntsev:
@perwendel 
* Yes, this will change the default behaviour. And that's the point. We must change the default behaviour! Current default behaviour creates a potential security holes in applications around the World. Currently `httpOnly` is not set by default, but it should be.
* User can add `httpOnly` **only to his own cookies** using `Response.cookie`. But user cannot add `httpOnly` to the **session cookie** (which is created by Jetty deep inside the framework).

Issue #1031, opened 2018-07-06
One staticFiles.externalLocation overwrites another


vladp:
Hello,

This is likely related to a similar issue
https://github.com/perwendel/spark/issues/1005
Only in my case, I am running into this leftover bug --with externalLocation.
externalLocations (at least in my case) cannot be subdirectories of each other  -- as that introduces a security risk/exploit. 
So I could not make one location as a subdirectory of another as @tipsy  recommend at some point to overcome this bug  for /public resource folders.

Details (using spark 0.7.2 )
---

I have a setup where a single java instance of Spark, is servicing 2 separate small web apps,
from 2 separate ports (of the same IP address).

I am leveraging the ignite method to setup to separate Services.

```
main () {

Service port4000 = Service.ignite();
port4000.port(4000);
port4000.staticFIles.location("/public");
port4000.staticFiles.externalLocation("/tmp/www/port4000app");
port4000.get (... )

Service port6000 = Service.ignite();
port6000.port(6000);
port6000.staticFIles.location("/public");
port6000.staticFiles.externalLocation("/tmp/www/port6000app");
port6000.get(...);
}
```
With the above configuration, I am getting an exception from DirectoryTravesal.java
```

    public static void protectAgainstForExternal(String path) {
        String nixLikePath = Paths.get(path).toAbsolutePath().toString().replace("\\", "/");
        if (!removeLeadingAndTrailingSlashesFrom(nixLikePath).startsWith(StaticFilesFolder.external())) {
            throw new DirectoryTraversalDetection("external");
        }
    }
```
Because when comparing the requested directory at port 4000, it is using
StaticFilesFolder.external()   , and that returns a directory associated with port 6000
(because it was setup last).

This is happening because, it appears, that Spark does not associate static folders with 
a particular Service instance, and instead, this is just simply a global value that can only be initialized
 once per application.

But then, this either seems to be a bug, or I am not understanding correctly how to initialize 2 separate 
Services within same main, using the ignite() function.

I scanned the docs but could not find guidance on this.

Presumably, this can also cause a security exploit, when a attacker can get access
to capabilities that should not be available from a given port/service (because 
**protectAgainstForExternal** would not protect)

Thank you in advance,

PS: static instance of the staticFiles configuration here:
https://github.com/perwendel/spark/blob/master/src/main/java/spark/staticfiles/StaticFilesConfiguration.java

tipsy:
It's the same bug as in the issue you linked to.

[...]

Issue #1005, opened 2018-04-25
Bad request on multiple Static file locations


RyanSusana:
Hello friends,

I have been using sparkjava for a while now and am developing a framework ([Elepy](https://github.com/RyanSusana/elepy) using your framework.

The problem that I am running into is that when I have multiple jars with static resources on different Services, I get a Bad Request 400 on all but the first staticfile location. The different Services are hosted on port 80 and 1337 and get instantiated with ignite()

This means that per JVM I can only have one staticfile location? It does recognize that the resource exists, otherwise it would give me a 404 instead.

I looked through the source code and it has something to do with DirectoryTraversalDetection.

tipsy:
Hi @RyanSusana. I'm guessing this is because `StaticFilesFolder` has static variables:

```
private static volatile String local;
private static volatile String external;
```

RyanSusana:
Are there any plans on changing this or must I resort to multiple JVM's?

RyanSusana:
Why would it break backwards compatability?

tipsy:
Any change in current behavior is really a change in backwards compatibility. I don't remember what the current behavior is, but based on reading the code it looks like you can do:

```kotlin
val serviceA = ignite().apply { staticFiles.location("/public") }
val serviceB = ignite()
```

Then do a `GET` for static files to `serviceB` . If this is the case, people will have to add `staticFiles.location("/public") ` to their second service too.

 I'll fire up IntelliJ and check.

tipsy:
Ok, so it didn't work without calling `staticFiles.`, so that's good. Also, I found some strange behavior: 

Works:
```java
Service serviceA = Service.ignite();
serviceA.port(4001);
serviceA.staticFiles.location("/public/pages");
serviceA.init();

serviceA.awaitInitialization();

Service serviceB = Service.ignite();
serviceB.staticFiles.location("/public");
serviceB.port(4002);
serviceB.init();
```

Bad request: 
```java
Service serviceA = Service.ignite();
serviceA.port(4001);
serviceA.staticFiles.location("/public");
serviceA.init();

serviceA.awaitInitialization();

Service serviceB = Service.ignite();
serviceB.staticFiles.location("/public/pages");
serviceB.port(4002);
serviceB.init();
```

This can be fixed as a bug then, I can't imagine anyone really relying on this behavior.

RyanSusana:
Thank you for the help!

perwendel:
Fixed by #1036

PR #980, opened 2018-02-23
added server-sent-events

mtzagkarakis:
I just add event sourcing capabilities (Server-Sent-Events).
I needed ES for my current project so I implement it.

PR #1036, opened 2018-07-18
Each service has own static file location

lukecmstevens:
Fixes issues #1031 and #1005 where multiple static file locations in different services would cause a DirectoryTraversal issue.

### DIFF
--- a/src/main/java/spark/resource/ClassPathResourceHandler.java
+++ b/src/main/java/spark/resource/ClassPathResourceHandler.java
@@ -80,7 +80,7 @@ public class ClassPathResourceHandler extends AbstractResourceHandler {
             }
 
             if (resource != null && resource.exists()) {
-                DirectoryTraversal.protectAgainstInClassPath(resource.getPath());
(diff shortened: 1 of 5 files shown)


vladp:
Thx for the fixes! hope it gets merged by @perwendel  (together with another functional ehn. https://github.com/perwendel/spark/pull/980 ) .


I have been struggling to work around the problem by spawning multiple instances.

vladp:
@lukecmstevens   
is it possible for me to reference your build from Gradle?
It does not look like any of the proposed pull requests are going to get merged in (at least within 3-4 month from the fix) -- so looking for a way forward.

lukecmstevens:
@vladp 
I don't use gradle myself so I can't really help, but this stackoverflow question might be able to https://stackoverflow.com/questions/18748436/is-it-possible-to-declare-git-repository-as-dependency-in-android-gradle

vladp:
thank you


________________________________

perwendel:
reviewing

PR #1055, opened 2018-09-18
Make Jetty Server simple to configure

sclassen:
Goal of this change is to simplify the configuration of the embedded Jetty server.

This is achieved by adding more configuration options to the EmbeddedJettyFactory.

`withHandlerBeforeRouteHandler(Handler handler)`
Allows to register Handlers which are executed before the Handler handling the spark routes.

`withJettyServerAttribute(String name, Object attribute)`
Allows to register attributes which are set onto a newly created jetty server instance.

`withJettyServerConfigurator(Consumer<Server> configurator)`
Allows to register a callback which will be triggered whenever a new jetty server instance is created.

### DIFF
--- a/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyFactory.java
+++ b/src/main/java/spark/embeddedserver/jetty/EmbeddedJettyFactory.java
@@ -16,6 +16,9 @@
  */
 package spark.embeddedserver.jetty;
 
+import org.eclipse.jetty.server.Handler;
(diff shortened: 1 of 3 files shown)


toyg:
I like the approach but it lacks the last bit of the trinity - connectors, which is where SSL parameters and other low-level parameters are set.

The main drawback of this approach is that it forces one to understand how Jetty works under the hood (what is a handler? etc), but it's probably the lesser evil.

PR #1071, opened 2018-10-17
Expose get/set on Service.embeddedServerIdentifier

crutcher:
Expose embeddeServerIdentifier, so that EmbeddedServers can be used to attach a different factory service to embedded Spark servers.


perwendel:
:+1: 
Thanks for contributing!

PR #1092, opened 2019-02-15
Update RouteEntry to support single wildcard.

intUnderflow:
This updates RouteEntry so that if a path is a single wildcard (ie `*`) it will be accepted as matching any string. This makes it possible to create a catch-all spark route, for example via:

```
spark.get("*", ...)
```


perwendel:
Thanks for commiting. I'll have a look at it, maybe next week. I find the functionality a bit strange but I might be wrong.

perwendel:
You have this functionality by writing the code:
```java
get("/*", (q, a) -> "Wild world!");
```

Issue #1006, opened 2018-04-27
Get all mapped routes


buckelieg:
Provide method to retrieve all mapped routes (in all route groups).

ErnestoSalazar:
Something like a method that list all mapped routes like this?
--------------------------------------------------
/hello -> before
/hello -> post
/hello -> after
--------------------------------------------------

ErnestoSalazar:
Something like a method that list all mapped routes like this?
-=============================
/hello -> before
/hello -> post
/hello -> after
-=============================

danieljamesscott:
I would also like this. Exposing Service.pathDeque would be great, even better if we could get the HTTP method too. Would you accept a PR which did this?

buckelieg:
The idea was not to acess routes as objects (but thi is interesting idea). Just qanted to have a list of all routes mapped. The use case of mine is that I want to create a component that restrincts access to particular route (URL) with dynamic configuratrion. The thing is if I have added the routes they are added to the list right away.

buckelieg:
Whenever I want to configure restrictions through UI - then yes. 
I am thinking about ONE security filter which uses configuration from DB which is in turn is configured from UI outside.

buckelieg:
Any chances to get this implemented?

johnnybigoode:
Heya, this use case seems necessary for me and I would like to try adding this feat. If any senior devs could give me some idea of the best practice for this, it would be greatly appreciated.

johnnybigoode:
@buckelieg 

https://github.com/perwendel/spark/pull/1097

johnnybigoode:
So we would like to retrieve a liste of all routes avaliable within an ignited service. 

So it makes sense to have a 'findAllRoutes' method in the Service class and another one in the Spark class, since the service is not clearly exposed from the Spark class. 

The Service class has a Routes field that seems to hold all Routes available in a service. And when we stop the server, it clears the Routes list, which makes sense. 

Whithin the Routes class we have two interesting methods 'find' and 'findMultiple' and both of them return a simple data class that stores 4 values:

    private Object target;
    private String matchUri;
    private String requestURI;
    private String acceptType;

More interestingly, there are two other private methods that find things, 'findTargetsForRequestedRoute' and 'findTargetWithGivenAcceptType', both of them return 'RouteEntry' objects. 

[...]

PR #1097, opened 2019-03-06
Add FindAll to Routes>Service>Spark>RoutesFindAllTest implements #1006

johnnybigoode:
So, from what I understood. The class `Spark` starts the whole thing, but you can create additional `Spark`s with the static method `spark.Service.ignite`

Which literally gives us a `Service` object.
The service class seems to have a field that holds all the routes called routes.

Exposing that information would be the way to go. 

In the Routes class, there are no methods for returning the whole route list, so that's where I started.

Using a similar structure to the method `findMultiple` in ln 124. I noticed that we could use the method `findTargetsForRequestedRoute` to return all, but it depends on `RounteEntry.matches` method that states:

ln 47:
```
        if (
                (httpMethod == HttpMethod.before || httpMethod == HttpMethod.after || httpMethod == HttpMethod.afterafter)
                && (this.httpMethod == httpMethod)
                && this.path.equals(SparkUtils.ALL_PATHS)) {
            // Is filter and matches all
            return true;
        }
```

It was expected to have an option to return all. But it's not possible to find an httpMethod that returns ALL options. Maybe a better implementation would be creating a `SparkUtils.ALL_METHODS` and using something on those lines to create a better functionality. 

In our case, this is not necessary. If getting all routes means exposing the routes available to the dev, is it unthinkable to find reason to make this method internal. Since the only thing available to change routes seems to be `Routes.clear();` and it is only used by `Service.initiateStop()`, it's safe to say that routes are not mutable, so a `findAll()` just needs to return a copy of all routes available to the running service. 

#tldr:

I'm adding a `Routes.findAll()` method. Takes no arguments and returns a new `List<RouteMatch>` with all that's available. Since there's not http method, path or accepted type to add to the `RouteMatch`, I'm just adding all the `routeEntry` information to the results. 

    `matchSet.add(new RouteMatch(routeEntry.target, routeEntry.path, routeEntry.path, routeEntry.acceptedType));`

The `Routes` is used within `Service`, so I added a findAll to Routes and a `findAllRoutes `to Service and a static `findAllRoutes `to Spark.

New test for these features: `RoutesFindAllTest`.

### DIFF
--- a/src/main/java/spark/QueryParamsMap.java
+++ b/src/main/java/spark/QueryParamsMap.java
@@ -9,14 +9,14 @@ import java.util.regex.Pattern;
 import javax.servlet.http.HttpServletRequest;
 
 /**
- * This objects represent the parameters sent on a Http Request. <br>
+ * These objects represent the parameters sent on a Http Request. <br>
  * Parses parameters keys like in Sinatra. <br>
  * <br>
  * For a querystring like: <br>
  * user[name]=federico&#38;user[lastname]=dayan
  * <br>
  * <br>
- * We get would get a structure like: <br>
+ * We get a structure like: <br>
  * user : {name: federico, lastname: dayan}
  * <br>
  * <br>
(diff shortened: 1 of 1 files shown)


buckelieg:
Wow. Great!
Just one thing to add - could the rersults return collection of RouteEntry objects which do contain HTTP method info:
In the `Service.class`:
```java
Collection<RouteEntry> routes();
```
If we map different HTTP methods to the same URL we could not set up the things correctly.
Thank a lot!

johnnybigoode:
> If we map different HTTP methods to the same URL we could not set up the things correctly.

While I understand your concern and your use case, I'm not sure how to go from here. From what I checked from the code, all RouteMatch situations require to pass a `HttpMethod` and it's not something I predicted. 

Gonna reviews this and I'll get in touch.

PR #1100, opened 2019-03-13
Get all mapped routes

johnnybigoode:
I want to implement #1006. 

From what I understood, adding a `httpRequest` field to `RouteMatch `and exposing a `getRoutes` method in the `Service` and `Spark `classes.

A few lines were changed to accommodate the new field. 

And a new test in `MultipleServicesTest` to check if everything is dandy.

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -36,6 +36,7 @@ import spark.embeddedserver.jetty.websocket.WebSocketHandlerWrapper;
 import spark.route.HttpMethod;
 import spark.route.Routes;
 import spark.route.ServletRoutes;
+import spark.routematch.RouteMatch;
 import spark.ssl.SslStores;
 import spark.staticfiles.MimeType;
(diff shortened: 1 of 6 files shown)


perwendel:
Thanks for commiting. I'll have a look next week.

perwendel:
Reviewed and tested!!
Thanks for contributing!

Paxa:
May be it will be useful (with spark 2.9.4)

```java
import spark.Spark;
import spark.routematch.RouteMatch;

// somewhere after route definitions

for (RouteMatch route : routes()) {
    String method = route.getHttpMethod().toString();
    if (method.equals("before") || method.equals("after")) continue;
    System.out.printf("--> %-7s %s\n", method.toUpperCase(), route.getMatchUri());
}

```

output
```
--> POST    /internal/update_notification
--> GET     /internal/history
--> GET     /internal/history_details
--> GET     /ping
```

PR #1137, opened 2019-09-03
Initiate the servlet instance in exception mapper

VithyaG:
Without this change the servlet instance will be initialised again when the matcher filter is initialised.


lepe:
PR merged into/solved in the [unofficial build - Release 2+](https://github.com/Intellisrc/spark/)

PR #1170, opened 2020-05-31
Fix issue 1072

xrl2408:


perwendel:
@xrl2408 Thanks for your contribution!

PR #1171, opened 2020-05-31
Fix issue 986 & 1146

WingChan1999:
- Fix issue 986: Web socket request gets handled by get route

- Fix issue 1146: Spark.webSocketIdleTimeoutMillis should handle Long instead of Integer


perwendel:
I did an initial review of this and it looks good.
There are some backwards compatibility issues I'll have to think about how to solve before merging it.

PR #1173, opened 2020-05-31
fix for issue 911

xjh33445566:
I just do some simple code to fullfill the feature request.
Hope this code can do some help for you.

### DIFF
--- a/src/main/java/spark/http/matching/Body.java
+++ b/src/main/java/spark/http/matching/Body.java
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package spark.http.matching;
 
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


lepe:
PR merged into/solved in the [unofficial build - Release 2+](https://github.com/Intellisrc/spark/)

AN3Orik:
Sorry, but this is bull**it. Why json type by default? That's broke any template engine integration out-of-box.

AN3Orik:
This commit also broke logic where application already have content-type redefentions for POST requests. For example: jQuery ajax parse result depend on Content-Type received from controller. + default auto detection content type by extension making to use this  fork impossible for projects where spark used more than just for hold REST requests.

Not to mention about sneaky Configuration outside Spark class.

PR #1184, opened 2020-08-13
Test improvement: removed Magic Number test smell

gvma:
This is a test refactoring

**Problem:**
The Magic Number Test occurs when assert() statements in a test method contain numeric literals (i.e., magic numbers) as parameters.

**Solution:**
As magic numbers do not indicate the meaning/purpose of the number, they should be replaced with constants or variables, thereby providing a descriptive name for the input.

_Before:_
`Assert.assertEquals(200, response.status);
`
_After:_
`Assert.assertEquals(HTTP_OK, response.status);`

### DIFF
--- a/src/test/java/spark/BodyAvailabilityTest.java
+++ b/src/test/java/spark/BodyAvailabilityTest.java
@@ -18,9 +18,11 @@ public class BodyAvailabilityTest {
     private static final Logger LOGGER = LoggerFactory.getLogger(BodyAvailabilityTest.class);
 
     private static final String BODY_CONTENT = "the body content";
-
+    
     private static SparkTestUtil testUtil;
 
+    private final int HTTP_OK = 200;
+    
     private static String beforeBody = null;
     private static String routeBody = null;
     private static String afterBody = null;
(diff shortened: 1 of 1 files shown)


src/test/java/spark/BodyAvailabilityTest.java
```diff
@@ -18,9 +18,11 @@
     private static final Logger LOGGER = LoggerFactory.getLogger(BodyAvailabilityTest.class);
 
     private static final String BODY_CONTENT = "the body content";
-
+    
     private static SparkTestUtil testUtil;
 
+    private final int HTTP_OK = 200;
+
```
Ariel-Isaacm, line 25: Maybe removing this extra line? Also I feel it would be better to group all the constant together?

gvma: Hello, Ariel!

Let me see if I get this straight... By grouping all the constant together you mean creating some sort of class for all HTTP status or do you only mean stacking up all the constants in lines close to each other?

Ariel-Isaacm: Hi! 

By grouping them together I was referring to just have them one next to another something like: 

```
    private static final Logger LOGGER = LoggerFactory.getLogger(BodyAvailabilityTest.class);
    private final int HTTP_OK = 200;
    private static final String BODY_CONTENT = "the body content";

    private static SparkTestUtil testUtil;


It's a minor thing tho 😅

PR #1193, opened 2020-09-14
Modify Base64 to use java.util.Base64 Encoder/Decoder rather than the sun.misc classes

rbygrave:
With Java 8 we can migrate to use `java.util.Base64.Decoder` / `Encoder` rather than  `sun.misc.BASE64Decoder` and `sun.misc.BASE64Encoder`.  

The  `sun.misc` classes are not be available in later versions of Java.

### DIFF
--- a/src/main/java/spark/Base64.java
+++ b/src/main/java/spark/Base64.java
@@ -16,15 +16,12 @@
  */
 package spark;
 
-import java.io.IOException;
-
-import sun.misc.BASE64Decoder;
-import sun.misc.BASE64Encoder;
-
-
 class Base64 {
 
     //CS304 Issue link:https://github.com/perwendel/spark/issues/1061
+
+    private static final java.util.Base64.Encoder urlEncoder = java.util.Base64.getUrlEncoder();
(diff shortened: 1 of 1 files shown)


src/main/java/spark/Base64.java
```diff
@@ -16,16 +16,13 @@
...
-
-import sun.misc.BASE64Decoder;
-import sun.misc.BASE64Encoder;
-
-
 class Base64 {
 
     //CS304 Issue link:https://github.com/perwendel/spark/issues/1061
 
+    private static final java.util.Base64.Encoder urlEncoder = java.util.Base64.getUrlEncoder();
```
perwendel, line 23: I don't know if this is valid for Java 8s Base64 class but I have a vague memory that it's faster not assigning an encoder/decoder to a static instance and then re-use for each decoding/encoding.

perwendel: maybe it had to do with not being threadsafe...hmmm. Do you know? I'll try to find out.

perwendel:
Reviewed and (performance) tested.

Issue #320, opened 2015-08-08
X-Forwarded-For support in request.ip()


angryziber:
Currently, when Sparks runs behind a proxy or other front-end server, request.ip() returns the IP of the proxy server, which is usually not what you want.

Most proxy servers set the X-Forwarded-For header with user's real IP, so request.ip() could take advantage of that.

PR #1197, opened 2020-10-18
added trustForwardHeaders method to decide if to trust HTTP forward headers

Arisstath:
It seems that from https://github.com/perwendel/spark/issues/320 Spark blindly trusts HTTP headers that can be spoofed, with no way to disable this behavior.

The newly introduced method ```Spark.trustForwardHeaders``` allows you to specify if you want to use jetty's ForwardedRequestCustomizer (https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/server/ForwardedRequestCustomizer.html)

### DIFF
--- a/src/main/java/spark/Service.java
+++ b/src/main/java/spark/Service.java
@@ -95,6 +95,8 @@ public final class Service extends Routable {
         System.exit(100);
     };
 
+    private boolean trustForwardHeaders = true;
+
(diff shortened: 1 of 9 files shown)


perwendel:
I made some modifications to maintain backwards compatibility and renamed some methods. Thanks for contributing.

PR #1223, opened 2021-04-24
Fix #1077 : Solve the bad route selection based on acceptType.

### LINKED ISSUE #1077: Bad route selection based on acceptType.

Chauncey-Xxy:
Try to fix #1077
I am not sure whether it is a bug, because it is tagged as Bug ..?, but I think it conflict with the documentation, so I try to fix it.

In short, we expect the input with empty acceptType will match any possible types as follows:
 curl -i -H "Accept: application/json" http://localhost:4567/hello : Hello application json
 curl -i -H "Accept: text/html" http://localhost:4567/hello : Go Away!!!
 curl http://localhost:4567/hello : Hello application json
Now the first and second are right, but the last command will get output: Go Away!!!
I think it is not reasonable because the empty acceptType should match every possibilities and use the earliest match principle, it should match the first possible acceptType.
So I change the codes to match "*/*" with the earliest acceptType.

Co-Authored-By: Bugjudger https://github.com/Bugjudger

### DIFF
--- a/src/main/java/spark/http/matching/MatcherFilter.java
+++ b/src/main/java/spark/http/matching/MatcherFilter.java
@@ -125,15 +125,15 @@ public class MatcherFilter implements Filter {
         HttpMethod httpMethod = HttpMethod.get(httpMethodStr);
 
         RouteContext context = RouteContext.create()
-            .withMatcher(routeMatcher)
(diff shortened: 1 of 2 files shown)


Chauncey-Xxy:
There are some mistakes and I will try it again for some reasons

PR #1230, opened 2021-05-18
fix-issue339 have ": " in request params map

nicetryzz:
I changed the getParams method in request method, in order to guarantee the match path doesn't start with ":".


nicetryzz:
Also, url in http should be Sensitive. Change it to low case isn't a good behavior.

PR #1234, opened 2021-05-20
Solve the problem of Method not allowed (405)

### LINKED ISSUE #1226: Method not allowed (405) but the response return not found (404)
With simple example at https://github.com/perwendel/spark/blob/master/README.md#examples
When I curl a request with the not declare method. Ex:
```
curl --location --request PUT 'localhost:4567/hello'
```
The response return status 404, but not is 405

Bugjudger:
Try to fix #1226 

When a website only can be visited by the GET method and you use the PUT method, you will get 405 error but not 404.

You can test it: curl -X PUT "localhost:4567"


lepe:
This and other PR have been merged into [intellisrc/spark](https://github.com/Intellisrc/spark/) fork (unofficial build)

PR #1257, opened 2022-06-17
Add test for Base64.decode and Base64.encode

lacinoire:
Hey 😊
I want to contribute a test.
Curious to hear what you think!

(I wrote this test as part of a research study at TU Delft. [Find out more](https://github.com/lacinoire/lacinoire/blob/main/README.md))


asolntsev (changes requested):
Usually every unit-test should verify only one method. I would split your test into two: one for encode and other for decode.

lacinoire:
Good point, I've split it into one test each.

PR #1280, opened 2023-07-21
fix(sec): upgrading jetty version to 9.4.51.v20230217

bowika:
Fixing CVE, this version can be pulled with jitpack.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
 
     <properties>
         <java.version>1.8</java.version>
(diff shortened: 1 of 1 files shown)


bowika:
I have just now put a commit on top for fixing further ones. Until it is merged, it may use it via jitpack with
eg. pom.xml:
```
        <dependency>
            <groupId>com.github.bowika</groupId>
            <artifactId>spark</artifactId>
            <version>40e9a4fc7e</version>
        </dependency>
```

